PR #4, opened 2024-08-29
Add newswires database

bryophyta:
## Accessibility

### DIFF
--- a/cdk/lib/__snapshots__/wires-feeds.test.ts.snap
+++ b/cdk/lib/__snapshots__/wires-feeds.test.ts.snap
@@ -8,6 +8,10 @@ exports[`The WiresFeeds stack matches the snapshot 1`] = `
       "GuS3Bucket",
       "GuDistributionBucketParameter",
       "GuLambdaFunction",
+      "GuVpcParameter",
+      "GuSubnetListParameter",
+      "GuSecurityGroup",
+      "GuDatabase",
     ],
     "gu:cdk:version": "TEST",
   },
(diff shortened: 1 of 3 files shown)


ingestion-lambda/src/handler.ts
```diff
@@ -1,83 +1,86 @@
...
+				body,
+			}): Promise<OperationResult> => {
+				const sqsMessageId = messageId;
 
-	const responses = await Promise.all(
-		eventBodies.map(
-			({ sqsMessageId: messageId, messageAttributes, snsMessageContent }) => {
-				const fingerpostMessageId =
-					messageAttributes['Message-Id']?.stringValue;
+				try {
```
bryophyta, line 31: `try {} catch {}` around the whole block doesn't feel _ideal_, but it felt like it made sense for now given that we're not actually sure yet how we want to handle the potential dependencies between the different operations

ingestion-lambda/package.json
```diff
@@ -3,13 +3,19 @@
...
 		"typecheck": "tsc -noEmit",
 		"build": "esbuild src/handler.ts --bundle --minify --outfile=dist/handler.js --external:aws-sdk --platform=node",
 		"test": "jest --detectOpenHandles --config ../jest.config.js --selectProjects ingestion-lambda"
 	},
 	"devDependencies": {
 		"@aws-sdk/client-s3": "3.637.0",
 		"@types/aws-lambda": "8.10.145"
+	},
+	"devDependencies": {
+		"@aws-sdk/client-rds": "3.637.0",
```
twrichards, line 16: not sure we still need this?

ingestion-lambda/package.json
```diff
@@ -3,13 +3,19 @@
 	"version": "1.0.0",
 	"description": "",
 	"scripts": {
-		"dev": "tsx watch src/handler.ts",
+		"dev": "tsx watch localRun.ts",
 		"typecheck": "tsc -noEmit",
 		"build": "esbuild src/handler.ts --bundle --minify --outfile=dist/handler.js --external:aws-sdk --platform=node",
```
twrichards, line 8: might need to add `--external:@aws-sdk ` to this bundle command? (to ensure we exclude everything aws-sdk)

ingestion-lambda/src/handler.ts
twrichards: could you change to [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) and allow the async block to throw (which gets turned into promise rejection) then iterate, to log any failures.

ingestion-lambda/src/handler.ts
```diff
@@ -1,83 +1,86 @@
...
-					});
+				return { sqsMessageId, status: 'success' };
 			},
 		),
 	);
 
-	const batchItemFailures = responses
-		.filter(({ status }) => status === 'rejected')
-		.map(({ messageId }) => ({ itemIdentifier: messageId }));
+	await sql.end();
```
twrichards, line 75: I would perhaps suggest wrapping as much as possible of this function in a `try` (optional `catch`), but with this `await sql.end()` in the `finally` block to ensure its always fired)

twrichards (commented):
looking really great, i tweaked the PR description slightly and added some comments, let me know when you need a re-review...

ingestion-lambda/package.json
bryophyta: in the build command there's already `--external:aws-sdk` -- do you mean we should replace this with `--external:@aws-sdk`, or have both?

ingestion-lambda/src/handler.ts
bryophyta: good call, thanks! done that

ingestion-lambda/package.json
twrichards: both, covers all bases

bryophyta:
https://github.com/guardian/newswires/pull/4#discussion_r1745386127

I've pushed a commit that logs the reason for failures 👍 

I'd be happy to use `Promise.allSettled` at some point for this, but not sure it's needed at the moment. Given that (afaik) we haven't settled on a desired behaviour for coping with failures yet, I'd be inclined to refactor later?

twrichards (approved):

PR #7, opened 2024-09-04
server connects to DB with IAM auth

andrew-nowak:
Set up the server with the ability and config to connect to the database with IAM auth. When running locally you can connect to the CODE db with an `ssm ssh` tunnel. There's currently no way to connect it to a local postgres, we'll need to figure out some way to switch between local<->local, and local<->CODE, but that's for the future.

Instructions for running with tunnel in the README.

### DIFF
--- a/newswires/app/conf/Database.scala
+++ b/newswires/app/conf/Database.scala
@@ -1,0 +1,86 @@
+package conf
+
+import play.api._
+import scalikejdbc._
+import software.amazon.awssdk.services.ssm.SsmClient
+import software.amazon.awssdk.services.ssm.model.GetParameterRequest
+import software.amazon.jdbc.ds.AwsWrapperDataSource
+
+import java.util.Properties
+
+object Database extends Logging {
+
+  def healthcheck = {
+    DB localTx { implicit session =>
+      sql""" SELECT 1 as successes """
+        .map(_.int("successes"))
+        .single()
+        .apply()
+        .get
+    }
+  }
+
+  // def getDevDbConfig(initialConfiguration: Configuration): Configuration = {
+  //   logger.info("building DB config for DEV")
+  //   val username = initialConfiguration
+  //     .get[String]("database.username")
+  //   val port =
+  //     initialConfiguration.get[Int]("database.port")
+  //   val hostname = initialConfiguration.get[String]("database.endpoint-address")
+  //   val password = initialConfiguration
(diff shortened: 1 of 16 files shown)


newswires/app/conf/Database.scala
```diff
@@ -0,0 +1,113 @@
...
+  //       "databaseName" -> "birthdays",
+  //       "password" -> password
+  //     )
+  //
+  //   Configuration.from(
+  //     Map(
+  //       "slick.dbs.default.db.properties" -> configuration
+  //     )
+  //   )
+  // }
```
twrichards, line 47: is this intentionally commented with the intention of reinstating soon for local postgres access (i.e. displaying what ingestion lambda is producing)

newswires/client/src/App.tsx
```diff
@@ -1,9 +1,229 @@
+import { css } from '@emotion/react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
```
twrichards, line 2: might as well remove `useCallback` as unused

newswires/client/src/App.tsx
```diff
@@ -1,9 +1,229 @@
...
+
+	const [query, setQuery] = useState<string>('');
+
+	const [selected, setSelected] = useState<WireData | undefined>(undefined);
+
+	const safeBodyText = useMemo(() => {
+		return selected?.content.body_text
+			? sanitizeHtml(selected.content.body_text)
+			: undefined;
+	}, [selected]);
```
twrichards, line 62: we probably ought to add quite a strict Content Security Policy (like we have for https://github.com/guardian/editorial-wires) before this hits production, to guard against malicious content (e.g. XSS) that might find its way into wire content (as it's presented on our domain). we should consider baking in the hash of our own JS bundle such that it's the only JS allowed to run on the page.

twrichards (approved):
nice work 🙌 
(couple of non-blocking comments)

newswires/app/conf/Database.scala
andrew-nowak: yep exactly

newswires/client/src/App.tsx
andrew-nowak: sounds good to me! That is exactly why I pass the content through `sanitize-html` first, but no reason not to do both - let's do this in a followup PR though

PR #8, opened 2024-09-06
split the app back up into well named stacks

andrew-nowak:
sqs/sns live in wires-feeds, everything downstream in newswires


twrichards (approved):
very sensible delineation

PR #10, opened 2024-09-06
lambda must be in VPC to get access to database

andrew-nowak:
## What does this change?

Lambda currently cannot access database, which makes sense because it doesn't have a security group granting it access.
Lambdas cannot be granted security group access unless they're in the VPC.
So add the lambda to the VPC.

## How to test

Deploy to CODE, can the lambda now write to the DB (or at least fail at a different point than "connect ETIMEDOUT")?

### DIFF
--- a/cdk/lib/newswires.ts
+++ b/cdk/lib/newswires.ts
@@ -1,11 +1,16 @@
-import type { GuStackProps } from '@guardian/cdk/lib/constructs/core';
+import { GuStackProps } from '@guardian/cdk/lib/constructs/core';
 import { GuCname } from '@guardian/cdk/lib/constructs/dns';
 import { GuParameter, GuStack } from '@guardian/cdk/lib/constructs/core';
 import { GuLambdaFunction } from '@guardian/cdk/lib/constructs/lambda';
 import { GuS3Bucket } from '@guardian/cdk/lib/constructs/s3';
 import type { App } from 'aws-cdk-lib';
 import { Duration } from 'aws-cdk-lib';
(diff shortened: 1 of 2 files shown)


cdk/lib/newswires.ts
```diff
@@ -71,9 +78,15 @@ export class Newswires extends GuStack {
...
 					DATABASE_NAME: databaseName,
 				},
+				vpc,
+				vpcSubnets: {
+					subnets: privateSubnets,
+				},
 			},
 		);
 
+		ingestionLambda.connections.allowTo(database, Port.tcp(5432));
```
twrichards, line 88: would the following not work?
```suggestion
	database.grantConnect(ingestionLambda);
```
(taken from an RDS Proxy example, so might legit not work, but if it does I think more readable)

twrichards: documentation here https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-rds.DatabaseInstance.html#grantwbrconnectgrantee implies it doesn't work though lol, although linked issue suggests it now does work https://github.com/aws/aws-cdk/issues/11851#issuecomment-1531639694

twrichards (approved):
LGTM but one suggestion (which may or may not work)

cdk/lib/newswires.ts
andrew-nowak: we have this in https://github.com/guardian/newswires/blob/71041b3923888dc23cdbd73e7a9604cae2aeaef0/cdk/lib/newswires.ts#L107 - `grantConnect` grants the IAM permission, but we also need to grant network access through security groups, which is what `connections.allowTo` grants

PR #13, opened 2024-09-11
Add start script and local db config

bryophyta:
## What does this change?

- Add a config function for connecting to the local postgres db
- Add a start script (modelled after [the Grid's `start.sh`](https://github.com/guardian/grid/blob/main/dev/script/start.sh) which:
  - Checks for necessary dependencies
  - By default will spin up the Docker containers, run the `ingestion-lambda` in 'dev' mode, and start the Play app.
  - If passed the `--use-CODE` flag, it will establish a tunnel to the CODE db, if there isn't one already, and start the Play app.

## Accessibility

### DIFF
--- a/newswires/app/AppComponents.scala
+++ b/newswires/app/AppComponents.scala
@@ -46,8 +46,11 @@ class AppComponents(context: Context)
     .build()
 
   if (context.environment.mode == Mode.Dev) {
-    // TODO run against a local DB?
-    Database.configureRemoteDevDb(ssmClient)
+    if (sys.env.contains("USE_CODE_DB")) {
+      Database.configureRemoteDevDb(ssmClient)
+    } else {
+      Database.configureLocalDevDB()
+    }
   } else if (context.environment.mode == Mode.Prod) {
     Database.configureDeployedDb(configuration)
   }
--- a/scripts/start
+++ b/scripts/start
@@ -18,7 +18,6 @@ ROOT_DIR=${DIR}/..
 # copied over from the Grid. Not 100% sure of the rationale, but perhaps: https://stackoverflow.com/questions/56520354/getting-an-amazonkinesisexception-status-code-502-when-using-localstack-from-ja
 export AWS_CBOR_DISABLE=true
 
-LOCAL_AUTH=false
 for arg in "$@"; do
   if [ "$arg" == "--debug" ]; then
     IS_DEBUG=true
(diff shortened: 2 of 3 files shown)


newswires/app/AppComponents.scala
```diff
@@ -46,8 +46,11 @@ class AppComponents(context: Context)
     .build()
 
   if (context.environment.mode == Mode.Dev) {
-    // TODO run against a local DB?
-    Database.configureRemoteDevDb(ssmClient)
+    if (sys.env.contains("USE_CODE_DB")) {
```
bryophyta, line 49: not super happy about using an environment variable here

scripts/start
```diff
@@ -0,0 +1,159 @@
...
+}
+
+tunnelToAwsDb() {
+    if (docker stats --no-stream &> /dev/null); then
+      docker compose down
+    fi
+    if [[ -n $EXISTING_TUNNELS ]]; then
+      echo "RE-USING EXISTING TUNNEL TO CODE POSTGRES (on port ${POSTGRES_PORT})"
+    else
+      # todo: Grid script uses the following options, do we need any of these?
```
bryophyta, line 82: I need to look into this still

scripts/start
```diff
@@ -0,0 +1,159 @@
...
+    hasCredentials
+    checkRequirements
+    checkNodeVersion
+    if [ $USE_CODE == true ]; then
+        echo "Using CODE. Starting play app with tunnel to AWS DB."
+        tunnelToAwsDb && USE_CODE_DB=true startPlayApp
+    else
+      echo "Using the local stack. Starting Docker, Ingestion Lambda and Play App."
+      # run all three, and terminate together when the user presses `ctrl+c`: https://stackoverflow.com/a/52033580
+      (trap 'kill 0' SIGINT; startDockerContainers && startIngestionLambda & startPlayApp & wait)
```
bryophyta, line 154: this seems to broadly work as expected, but I've noticed a couple of bugs with my shell when subsequently doing interactive stuff in the same shell session, so I think this needs a bit more investigation..

newswires/app/AppComponents.scala
twrichards: we discussed in-person and ideally we might instead have an environment variable more specific to connect differences, such as `USE_IAM_AUTH` but let's save for a future PR

scripts/start
twrichards: I've had trouble with backgrounding via `&` before, let's see how painful it gets when using day-to-day and improve in a future PR if needs be

twrichards (approved):

twrichards:
Awesome work @bryophyta this will make things much easier while we're developing 🙌

PR #14, opened 2024-09-11
Pin postgres version in local and RDS

bryophyta:
## Accessibility

### DIFF
--- a/cdk/lib/__snapshots__/newswires.test.ts.snap
+++ b/cdk/lib/__snapshots__/newswires.test.ts.snap
@@ -1288,6 +1288,7 @@ exports[`The Newswires stack matches the snapshot 1`] = `
         "DeletionProtection": true,
         "EnableIAMDatabaseAuthentication": true,
         "Engine": "postgres",
+        "EngineVersion": "16.4",
         "MasterUserPassword": {
           "Fn::Join": [
             "",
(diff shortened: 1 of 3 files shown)


twrichards (approved):
good to keep the local version in line with the version in AWS to avoid pernicious issues - the downside to pinning is that we need to manually upgrade Postgres periodically, but RDS supports automatic upgrades for both minor and major versions (sep. settings).

I'll approve but leave it up to you to decide on that tradeoff, potentially changing this PR to turn on auto upgrades (potentially just minor ones) but keeping the change to `docker-compose.yml` to match current version in AWS RDS.

bryophyta:
> good to keep the local version in line with the version in AWS to avoid pernicious issues - the downside to pinning is that we need to manually upgrade Postgres periodically, but RDS supports automatic upgrades for both minor and major versions (sep. settings).
> 
> I'll approve but leave it up to you to decide on that tradeoff, potentially changing this PR to turn on auto upgrades (potentially just minor ones) but keeping the change to `docker-compose.yml` to match current version in AWS RDS.

Good point, thanks! IMO we don't want the major version to be upgraded automatically? But minor versions seems useful. How does this look? https://github.com/guardian/newswires/pull/14/commits/3e7753900a2e2a28423cfdf4fa5604c7187b585e

andrew-nowak (approved):

PR #15, opened 2024-09-12
Configure `pg_trgm` extension in local db

bryophyta:
## What does this change?

We depend on the `pg_trgm` extension which needs to be enabled for a given db. Adding the command in the docker-compose file will enable it.

## How to test

- Run start script and then do a search in the UI. The request uses a `pg_trgm` operator so if it succeeds then the extension has been enabled.
- Dev-only, so no implications for CODE/PROD

## Accessibility

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,3 +23,5 @@ services:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=postgres
       - POSTGRES_DB=newswires
+    # enable pg_trgm extension for text search
+    command: ['postgres', '-c', 'shared_preload_libraries=pg_trgm']


twrichards (approved):
is this enabled by default in RDS? or has been turned on manually or via CDK?

bryophyta:
> is this enabled by default in RDS? or has been turned on manually or via CDK?

I haven't been able to find out whether it's enabled by default (it's listed as a [supported extension](https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-extensions.html#postgresql-extensions-16x) but idk if that means it's actually installed by default). I haven't found any mention of it in the CDK output or any of our other code.

Probably best to check with @andrew-nowak?

PR #19, opened 2024-09-18
start codifying db schema with migrations

andrew-nowak:
start a directory of DB migrations, so the schema is version controlled and can be iterated on and applied to different environments safely.

I've struggled and generally been dissatisfied with play evolutions in previous projects, as it's generally designed to work with runtime dependency injection rather than our preferred usage of compile-time/explicit injection, so I'm going back to flyway; somewhat annoying that it's a new dependency, but it can be run and applied independently of the main app, and is much easier to configure, which I find a positive, and generally doesn't try to lead you down the minefield of building "down" migrations

### DIFF
--- a/db/README.md
+++ b/db/README.md
@@ -2,10 +2,20 @@
 
 This directory contains database configuration, including and especially the database migrations.
 
-Run the migrations with [flyway](https://documentation.red-gate.com/flyway/flyway-cli-and-api/welcome-to-flyway), which can be installed with Homebrew.
+Run the migrations with a scala-cli script, `flyway.sc`, which wraps the [flyway](https://documentation.red-gate.com/flyway/flyway-cli-and-api/welcome-to-flyway)
+API. Install scala-cli with Homebrew:
 
-This directory includes config for connecting to the local database, and is run as part of the start script.
+```sh
+brew install Virtuslab/scala-cli/scala-cli
+```
 
-`flyway migrate`
+Then you can inspect the state of the database, and run any pending migrations, by running the script directly.
 
-Instructions for connecting to our pre-production and production databases are contained elsewhere.
+```sh
+./flyway.sc info local      # to see which migrations have been applied
+./flyway.sc migrate local   # to run all pending migrations
+```
+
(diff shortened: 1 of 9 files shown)


db/README.md
```diff
@@ -0,0 +1,11 @@
+# db
+
+This directory contains database configuration, including and especially the database migrations.
+
+Run the migrations with [flyway](https://documentation.red-gate.com/flyway/flyway-cli-and-api/welcome-to-flyway), which can be installed with Homebrew.
+
+This directory includes config for connecting to the local database, and is run as part of the start script.
+
+`flyway migrate`
```
bryophyta, line 9: I'm not quite clear about this -- is it saying that we should run `flyway migrate` directly? If so, under what circumstances?

db/README.md
```diff
@@ -0,0 +1,11 @@
+# db
+
+This directory contains database configuration, including and especially the database migrations.
+
+Run the migrations with [flyway](https://documentation.red-gate.com/flyway/flyway-cli-and-api/welcome-to-flyway), which can be installed with Homebrew.
+
+This directory includes config for connecting to the local database, and is run as part of the start script.
```
twrichards, line 7:
```suggestion
This directory includes config for connecting to the local database, and is run as part of the local dev start script.
```

bryophyta (approved):
one small comment about docs but I've run this locally (after deleting my previous local Docker container) and it all seems to work as expected 👍

db/migrations/V2__trigram_indices.sql
```diff
@@ -0,0 +1,18 @@
...
+
+CREATE INDEX headline_trgm_index ON fingerpost_wire_entry
+  USING GIN ((content->>'headline') gin_trgm_ops);
+
+CREATE INDEX subhead_trgm_index ON fingerpost_wire_entry
+  USING GIN ((content->>'subhead') gin_trgm_ops);
+
+CREATE INDEX keywords_trgm_index ON fingerpost_wire_entry
+  USING GIN ((content->>'keywords') gin_trgm_ops);
+
```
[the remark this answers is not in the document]
andrew-nowak: We tried this on the CODE database, somewhat surprisingly (at least to me) this index and query operator continues to work even on the JSON array!

PR #22, opened 2024-09-23
temporarily set the RDS minor version to 16.4, until RDS defaults version 16 to 16.4 or higher, when we can re-enable autoMinorVersionUpgrade

andrew-nowak:
As on the title

Currently Cloudformation refuses to apply our stack definition, because version "16" currently defaults to "16.3", and our CODE instance is already running "16.4", and RDS cannot perform a downgrade.

We can revert this once "16" defaults to "16.4", and then go back to taking automatic minor version upgrades.


bryophyta (approved):
[Worked in CODE](https://riffraff.gutools.co.uk/deployment/view/8d97b402-775d-481e-b31d-af273c0aaee7), lgtm!

PR #24, opened 2024-09-24
update start script required dependencies, and install scala-cli in setup script

andrew-nowak:
This requirement is from an earlier iteration of #19 which used flyway directly, rather than through scala-cli. Update the dependencies and the setup script to account for the changes.


bryophyta (approved):
👍👍

PR #28, opened 2024-09-26
add the tsvector search column and indices

andrew-nowak:
Adds a brand new, generated tsvector column to all entries, which builds vector for the interesting text fields in the JSON data, then updates the query method to use that to search all the text fields together, and gets rid of the (pretty poor) trigram search

### DIFF
--- a/newswires/app/db/FingerpostWireEntry.scala
+++ b/newswires/app/db/FingerpostWireEntry.scala
@@ -35,7 +35,8 @@ object FingerpostWireEntry extends SQLSyntaxSupport[FingerpostWireEntry] {
   implicit val format: OFormat[FingerpostWireEntry] =
     Json.format[FingerpostWireEntry]
 
-  override val columns = Seq("id", "external_id", "ingested_at", "content")
+  override val columns =
+    Seq("id", "external_id", "ingested_at", "content", "combined_textsearch")
   val syn = this.syntax("fm")
 
   def apply(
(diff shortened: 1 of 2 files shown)


newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -35,7 +35,8 @@ object FingerpostWireEntry extends SQLSyntaxSupport[FingerpostWireEntry] {
   implicit val format: OFormat[FingerpostWireEntry] =
     Json.format[FingerpostWireEntry]
 
-  override val columns = Seq("id", "external_id", "ingested_at", "content")
+  override val columns =
```
twrichards, line 38: this doesn't have the side-effect that the generated column is returned from the DB for each row does it?

twrichards (approved):
wonderful 👏 

one question, hopefully answer is no

newswires/app/db/FingerpostWireEntry.scala
andrew-nowak: ah good point! not this line, but the fact that there's a new column in the table will affect any `select *`s yep. we should update them to only select the columns we actually want...

bryophyta (approved):
lgtm! 🥳

newswires/app/db/FingerpostWireEntry.scala
andrew-nowak: fixed

PR #29, opened 2024-09-26
always split stringy keywords, even if empty string

andrew-nowak:
## What does this change?

empty strings are falsy, so messages with `keywords: ""` escaped being split into an array, but remained as strings so broke the deserializer in scala-land.

instead, always run split if a split function exists (ie. keywords is `string`), and default to an empty array (so the resulting entry can be parsed)


twrichards (approved):
good catch - thanks

PR #31, opened 2024-09-26
Log-shipping fix: adopt crosswordv2 log settings

bryophyta:
## What does this change?

Copies the logback config that we're using in [crosswordv2](https://github.com/guardian/crosswordv2/commit/ad3cd755a1b04e94936fec2c18f1574b66cfb2d1)

It includes a number of changes to `build.sbt` as well. It might be that not all of these changes are necessary but I don't understand the jackson eviction stuff well enough to know what's needed and what's not.

Initially created because we thought that the broken log shipping was due to our config, though have since realised that there was an upstream issue with logging. Nevertheless, could be a QoL improvement to have nicer logs?

## Accessibility

### DIFF
--- a/newswires/conf/logback.xml
+++ b/newswires/conf/logback.xml
@@ -1,50 +1,50 @@
-<?xml version="1.0" encoding="UTF-8" ?>
-
 <!-- https://www.playframework.com/documentation/latest/SettingsLogger -->
-
-<!DOCTYPE configuration>
-
 <configuration>
-  <import class="ch.qos.logback.classic.encoder.PatternLayoutEncoder"/>
-  <import class="ch.qos.logback.classic.AsyncAppender"/>
-  <import class="ch.qos.logback.core.FileAppender"/>
-  <import class="ch.qos.logback.core.ConsoleAppender"/>
 
-  <appender name="FILE" class="FileAppender">
+  <appender name="FILE" class="ch.qos.logback.core.FileAppender">
     <file>${application.home:-.}/logs/application.log</file>
-    <encoder class="PatternLayoutEncoder">
+    <encoder>
       <charset>UTF-8</charset>
-      <pattern>%d{yyyy-MM-dd HH:mm:ss} %highlight(%-5level) %cyan(%logger{36}) %magenta(%X{pekkoSource}) %msg%n</pattern>
+      <pattern>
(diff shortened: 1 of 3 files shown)


twrichards:
do we know if this ensures log output is JSON? (so it gets parsed in a structured way by ELK)

similarly does it support stack traces (currently they come out as separate lines, which makes debugging v. painful)

bryophyta:
> do we know if this ensures log output is JSON? (so it gets parsed in a structured way by ELK)
> 
> similarly does it support stack traces (currently they come out as separate lines, which makes debugging v. painful)

If I'm honest I don't understand all the wizardry which @andrew-nowak put into this for crosswords, but in my experience it's worked really nicely there. An example record from crosswordv2 with separate `message` and `stackTrace` fields can be seen here: https://logs.gutools.co.uk/s/editorial-tools/app/discover#/doc/67604290-baa6-11e9-bea2-633437abb232/logstash-ed-tools-2024.09.30?id=T0ajQpIBvaA8WQZa0ffv

newswires/conf/logback.xml
```diff
@@ -1,50 +1,50 @@
...
-         which otherwise risk being sent ANSI escape sequences that they cannot interpret.
-         See https://logback.qos.ch/manual/layouts.html#coloring
-    -->
-    <!-- <withJansi>true</withJansi> -->
-    <encoder class="PatternLayoutEncoder">
-      <charset>UTF-8</charset>
-      <pattern>%d{yyyy-MM-dd HH:mm:ss} %highlight(%-5level) %cyan(%logger{36}) %magenta(%X{pekkoSource}) %msg%n</pattern>
-    </encoder>
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <if condition='property("STAGE").contains("DEV")'>
```
andrew-nowak, line 15: add the .sbtopts file to make this work https://github.com/guardian/crosswordv2/blob/main/.sbtopts (this condition turns _off_ json encoding for running locally, which is much more human readable)

bryophyta: ty! https://github.com/guardian/newswires/pull/31/commits/93a41c8dba21516e4ca5a945a6155371e58d4d07

andrew-nowak (approved):

PR #33, opened 2024-09-26
add `beforeId` query param to `/api/search` and remove offset param to facilitate id-based pagination (e.g. for infinite scroll)

twrichards:
Prior to this PR our 'paging' was based on numerical offset, which can be problematic if the underlying data is changing (which it is as incoming items on the feed, may match the current search), more sensible is using the ID. Fortunately our ID is a `SERIAL` (so sequential). So this PR adds a `beforeId` query param to fetch X records with ids less than the on specified.

This can be used to facilitate infinite scroll by passing the ID of the last search result from the 'previous page' to fetch the next X older items, matching the current search.

### DIFF
--- a/newswires/app/db/FingerpostWireEntry.scala
+++ b/newswires/app/db/FingerpostWireEntry.scala
@@ -81,14 +81,12 @@ object FingerpostWireEntry extends SQLSyntaxSupport[FingerpostWireEntry] {
   def query(
       maybeFreeTextQuery: Option[String],
       maybeKeywords: Option[List[String]],
-      page: Int = 0,
+      maybeBeforeId: Option[Int],
       pageSize: Int = 250
   ): QueryResponse = DB readOnly { implicit session =>
-    val effectivePage = clamp(0, page, 10)
     val effectivePageSize = clamp(0, pageSize, 250)
-    val position = effectivePage * effectivePageSize
 
-    val whereClauses = List(
+    val commonWhereClauses = List(
       maybeKeywords.map(keywords =>
         sqls"""(${FingerpostWireEntry.syn.column(
             "content"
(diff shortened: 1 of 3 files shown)


newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -99,7 +97,13 @@ object FingerpostWireEntry extends SQLSyntaxSupport[FingerpostWireEntry] {
       )
     ).flatten
 
-    val whereClause = whereClauses match {
+    val dataOnlyWhereClauses = List(
```
bryophyta, line 100: It seems like a deliberate choice to construct this as a separate val before merging the two lists of clauses, but I can't work out why from the naming.. What am I missing?

twrichards: ahh, we don't pass the id based clauses on the keywordCounts bit, because keywords bit is all about the total for each keyword in the whole result set, so if we're polling (more #34 than this PR) we want those counts to be 'latest'

newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -109,14 +113,13 @@ object FingerpostWireEntry extends SQLSyntaxSupport[FingerpostWireEntry] {
...
                         | LIMIT $effectivePageSize
-                        | OFFSET $position
                         | """.stripMargin
       .map(FingerpostWireEntry(syn.resultName))
       .list()
       .apply()
 
     val keywordCounts = getKeywords(additionalWhereClauses =
-      whereClauses
+      commonWhereClauses
```
bryophyta, line 122: should we be applying the 'beforeId' clause to the keyword counts query as well?

I can see an argument either way..

bryophyta (approved):
lgtm!

newswires/app/db/FingerpostWireEntry.scala
twrichards: no I don't think so, I feel keyword counts ought to always be latest counts given the search terms (excluding `beforeId` and `sinceId`) I can picture this being the easiest on the client

PR #34, opened 2024-09-26
add `sinceId` query param to `/api/search` to ultimately facilitate polling

twrichards:
To complement https://github.com/guardian/newswires/pull/33 (which introduced `beforeId` query param, mainly for infinite scroll purposes), here we add a `sinceId` query param to fetch only items since the most recent one that we're already aware of. This should be useful for polling (efficiently).


bryophyta (approved):
lgtm!

PR #36, opened 2024-09-30
don't parse `firstVersion`, `versionCreated` and `dateTimeSent` content fields as `ZonedDateTime`, keep as String instead

twrichards:
...since there are a number of records in the DB which fail to parse PLUS we're currently not using these at all (neither server-side nor client-side)


bryophyta (commented):
👍

andrew-nowak (approved):

PR #38, opened 2024-09-30
Add basic timestamp to wire card

bryophyta:
## What does this change?

* Refactor the Feed component to use a table instead of a list of cards, to represent wire items.
* Build on this to add a timestamp to the table entries.

Trying to build effective formatting into the `EuiCard` component meant pushing against the grain of the framework, whereas a table layout facilitates it pretty well. Using the lower-level building blocks in Eui, rather than their opinionated `EuiBasicTable` component, meant we could get a pretty simple layout that works fine, imo.

**nb.** More of an opportunity than a regression: Eui's table component auto-scrolls to whichever row has focus, which is quite nice. We might want to rework keyboard support (including focus order etc.) later to see if we can get away with just leveraging Eui builtins?

## Images

image

## Accessibility


newswires/client/src/WireDetail.tsx
```diff
@@ -19,50 +22,67 @@ export const WireDetail = ({ wire }: { wire: WireData }) => {
...
 	return (
 		<Fragment>
-			<h3
-				css={css`
-					font-weight: 300;
-				`}
-			>
-				{wire.content.subhead}
-			</h3>
+			<EuiScreenReaderLive focusRegionOnTextChange>
```
bryophyta, line 32: this means that the Item box gets focused every time a new item is selected, which feels like the right behaviour for screenreaders and for general keyboard support?

newswires/client/src/WireDetail.tsx
```diff
@@ -19,50 +22,67 @@ export const WireDetail = ({ wire }: { wire: WireData }) => {
...
+					`}
+				>
+					{wire.content.subhead}
+				</h3>
+			</EuiScreenReaderLive>
 
 			<EuiSpacer size="m" />
 
-			<EuiDescriptionList listItems={listItems} />
+			<EuiDescriptionList>
```
bryophyta, line 44: pass-by refactor -- now we're getting a lot more wires we're getting quite a few entries that are missing fields, or which have `keywords: [""]`, so doing some checks before rendering empty headings makes sense.

can split out into a separate PR if wished

bryophyta:
closing in favour of separate PRs. For the main feature, see https://github.com/guardian/newswires/pull/40

PR #40, opened 2024-10-01
Add timestamps to feed

bryophyta:
## What does this change?

* Refactor the Feed component to use a table instead of a list of cards, to represent wire items.
* Build on this to add a timestamp to the table entries.

Trying to build effective formatting into the EuiCard component meant pushing against the grain of the framework, whereas a table layout facilitates it pretty well. Using the lower-level building blocks in Eui, rather than their opinionated EuiBasicTable component, meant we could get a pretty simple layout that works fine, imo.

**nb.** More of an opportunity than a regression: Eui's table component auto-scrolls to whichever row has focus, which is quite nice. We might want to rework keyboard support (including focus order etc.) later to see if we can get away with just leveraging Eui builtins?

## Images

image

## Accessibility

### DIFF
--- a/newswires/client/src/WireDetail.tsx
+++ b/newswires/client/src/WireDetail.tsx
@@ -3,6 +3,7 @@ import {
 	EuiDescriptionList,
 	EuiFlexGroup,
 	EuiFlexItem,
+	EuiScreenReaderLive,
 	EuiSpacer,
 } from '@elastic/eui';
 import { css } from '@emotion/react';
@@ -52,13 +53,15 @@ export const WireDetail = ({ wire }: { wire: WireData }) => {
 
 	return (
 		<Fragment>
-			<h3
-				css={css`
-					font-weight: 300;
-				`}
-			>
-				{wire.content.subhead}
-			</h3>
+			<EuiScreenReaderLive focusRegionOnTextChange>
+				<h3
+					css={css`
+						font-weight: 300;
+					`}
+				>
+					{wire.content.subhead}
+				</h3>
+			</EuiScreenReaderLive>
 
 			<EuiSpacer size="m" />
 
--- a/newswires/client/src/WireItemTable.tsx
+++ b/newswires/client/src/WireItemTable.tsx
@@ -2,6 +2,7 @@ import {
 	EuiFlexGroup,
 	euiScreenReaderOnly,
 	EuiTable,
+	EuiTableBody,
 	EuiTableHeader,
 	EuiTableHeaderCell,
 	EuiTableRow,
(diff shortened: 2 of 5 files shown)


newswires/client/src/WireDetail.tsx
```diff
@@ -52,13 +53,15 @@ export const WireDetail = ({ wire }: { wire: WireData }) => {
...
 	return (
 		<Fragment>
-			<h3
-				css={css`
-					font-weight: 300;
-				`}
-			>
-				{wire.content.subhead}
-			</h3>
+			<EuiScreenReaderLive focusRegionOnTextChange>
```
bryophyta, line 56: this means that the Item box gets focused every time a new item is selected, which feels like the right behaviour for screenreaders and for general keyboard support?

newswires/client/src/WireItemTable.tsx
```diff
@@ -0,0 +1,125 @@
+import {
```
bryophyta, line 1: this is a big diff in the Github UI, due to the file being renamed, but I did the renaming in a separate commit so looking at the first commit will hopefully give a nicer diff of the main substantive changes

newswires/client/src/WiresCards.tsx
```diff
@@ -21,62 +28,98 @@ export const WireCardList = ({ wires }: { wires: WireData[] }) => {
...
+				<EuiTableHeader
+					css={css`
+						${euiScreenReaderOnly()}
+					`}
+				>
+					<EuiTableHeaderCell>Headline</EuiTableHeaderCell>
+					<EuiTableHeaderCell>Version Created</EuiTableHeaderCell>
+				</EuiTableHeader>
+				{wires.map(({ id, content }) => (
+					<WireDataRow
```
andrew-nowak: there are console warnings `Warning: validateDOMNesting(...): <tr> cannot appear as a child of <table>. Add a <tbody>, <thead> or <tfoot> to your code to match the DOM tree generated by the browser` - Is there a `EuiTableBody` component or something similar that needs to be added here?

newswires/client/src/WireDetail.tsx
andrew-nowak: Sounds right for screenreaders, but I think I could argue the opposite for "general" keyboard support - if I was keyboard controlling I think most handy would be a way to "peek" into stories as I scroll through them without moving my focus, so I could then quickly move onto the next in the story. Might be worthy of more nuanced keyboard control than just tab-support though, so possibly a future improvement

andrew-nowak (approved):

newswires/client/src/WireDetail.tsx
bryophyta: > if I was keyboard controlling I think most handy would be a way to "peek" into stories as I scroll through them without moving my focus, so I could then quickly move onto the next in the story. Might be worthy of more nuanced keyboard control than just tab-support though, so possibly a future improvement

I think I basically agree with you. Initially I was thinking that it was okay because we'd encourage using arrow keys for switching between items, rather than tab keys. However I've been rethinking that a bit. But agree that we probably want to have a more 'holistic' review of the keyboard behaviour later once we've got a clearer sense of the overall shape of the app?

newswires/client/src/WiresCards.tsx
bryophyta: good call, thanks! I'd checked the DOM structure (which was valid) but didn't realise that Eui had this more stringent requirement. The docs aren't great but as you suggested there is indeed an `EuiTableBody` component and adding it seems to have got rid of the warning 👍👍

PR #43, opened 2024-10-02
strip badly escaped curly quotes when necessary

andrew-nowak:
## What does this change?

Some stories are coming in with curly quotes escaped which isn't a valid JSON escape sequence as curly quotes don't need escaping. Detect, locate and remove these unnecessary escape sequences, eg `"\“abc\”"` -> `"“abc”"`

## How to test

There's a few messages of this sort on the DLQ, we can redrive those once this is merged.

Fewer missing messages

### DIFF
--- a/ingestion-lambda/src/handler.ts
+++ b/ingestion-lambda/src/handler.ts
@@ -28,6 +28,17 @@ const safeBodyParse = (body: string): any => {
 	} catch (e) {
 		if (e instanceof SyntaxError && isCurlyQuoteFailure(e)) {
 			console.warn('Stripping badly escaped curly quote');
+			// näive regex to delete a backslash before a curly quote - isn't a fully correct solution.
+			// what if a string had 2 backslashes then a curly quote? that would be a correct string, that
+			// we'd make incorrect by deleting the second backslash. We could fix that, but then what about
+			// 3 backslashes then a curly quote? etc. infinitely
+			// Generally it seems pretty unlikely that text would include literal backslashes before curly quotes,
+			// if we start finding some then we should investigate improvements...
+			//
+			// I got to this monstrosity:
+			//   .replaceAll(/(?<![^\\]\\(?:\\\\)*)\\(?!["\\/bfnrt]|u[0-9A-Fa-f]{4})/g, '')
+			//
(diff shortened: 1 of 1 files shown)


paperboyo (approved):
I feel like every curly quote anywhere is my fault, so I approve of this effort! (blindly, though!)

ingestion-lambda/src/handler.ts
```diff
@@ -18,6 +18,22 @@ interface OperationSuccess {
...
+	return !!e.message.match(/Unexpected token '[“‘”’]'/);
+};
+
+const safeBodyParse = (body: string): any => {
+	try {
+		return JSON.parse(body);
+	} catch (e) {
+		if (e instanceof SyntaxError && isCurlyQuoteFailure(e)) {
+			console.warn('Stripping badly escaped curly quote');
+			return JSON.parse(body.replaceAll(/\\([“‘”’])/g, '$1'));
```
itsibitzi, line 31: I think this might cause issues if there's an escaped backslash before a quote?

Maybe we want to make sure there isn't a `\\` in front of the quotes? e.g. `(^\\|[^\\]\\)[“‘”’]`.

https://regex101.com/r/r0V6ab/1

andrew-nowak: good point! I was feeling lazy, as literal backslashes are unlikely to appear in prose, especially near curly quotes...
Plus if we want to follow that rabbit hole, we then would need to consider if there was an escaped backslash and _then_ a badly-escaped curly quote, ie. `\\\“` should become `\\“`...

I think this is negative lookbehind territory, I got to https://regex101.com/r/6n6oNh/1 which seems to be behaving fairly well but looks pretty monstrous

itsibitzi: It's a fair point, maybe this concern is just WONTFIX.

andrew-nowak: added a comment explaining why this is only a partial fix, and some next steps if the problem rears its head again

bryophyta (approved):
good catch! and this looks good to me; feels like the proposed solution is a step in the right direction and strikes a good pragmatic balance 👍

PR #46, opened 2024-10-03
Add the request logging filter

andrew-nowak:
## What does this change?

Brings across my favourite copy/pasted code: the request logging filter

Adds logs that output an access-log-style entry for each request, eg. `127.0.0.1 - "GET /api/item/1 HTTP/1.1" 200 0 "https://newswires.local.dev-gutools.co.uk/item/1" 3ms`

## How to test

Run locally or on CODE, watching logs as you click around

## How to measure success

We can now observe what is happening in logs, to diagnose crashes etc.


bryophyta (approved):
lgtm!

PR #48, opened 2024-10-03
Set max concurrency for ingestor

andrew-nowak:
## What does this change?

When dumping lots of messages onto the queues at once (eg. redriving the DLQ) we dont want the lambda to spin up hundreds of simultaneous executions... this will overwhelm the database which is bad!

Sets the max concurrency to 10, and since each execution can handle up to 10 messages, this is up-to 100 messages inflight simultaneously which should be more than enough even for redrives!


bryophyta (approved):
Seems sound to me!

From the volume of activity we've seen on the wires so far it seems very unlikely that this would ever become a significant bottleneck for getting content onto the UI, but I suppose we should try to keep it in mind as a possible factor if we ever set up SLAs for 'time from SNS to UI', or similar.

PR #49, opened 2024-10-07
add some more fields into the FingerpostWire case class

andrew-nowak:
also do some field renaming to make life easier

## What does this change?

There's more subfields on the wires data coming through now; parse and surface the most interesting of these in the API.
There are more still to discover and interrogate, but I'm less certain that these will contain interesting values for us.

## How to test

Deploy to CODE and check API responses.

### DIFF
--- a/newswires/app/db/FingerpostWireEntry.scala
+++ b/newswires/app/db/FingerpostWireEntry.scala
@@ -1,27 +1,59 @@
 package db
 
-import play.api.libs.json.{Json, OFormat, OWrites}
+import play.api.libs.json._
 import scalikejdbc._
 
 import java.time.ZonedDateTime
 
+case class FingerpostWireSubjects(
+    code: List[String]
+)
+object FingerpostWireSubjects {
+  implicit val format: OFormat[FingerpostWireSubjects] =
+    Json.format[FingerpostWireSubjects]
+}
+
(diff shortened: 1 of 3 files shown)


newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -1,27 +1,59 @@
...
+  // rename a couple of fields
+  private val reads: Reads[FingerpostWire] =
+    Json.reads[FingerpostWire].preprocess { case JsObject(underlying) =>
+      JsObject(underlying.map {
+        case ("source-feed", v) => ("sourceFeed", v)
+        case ("body_text", v)   => ("bodyText", v)
+        case pair               => pair
+      })
+    }
+  //// more play-jsony alternative:
```
bryophyta, line 47: fwiw I prefer the version that you've got -- feels more readable to me at least

should we remove the commented code or do you think it's useful for documentation?

bryophyta (approved):
this lgtm! and the renamings seem clear & sensible to me

newswires/app/db/FingerpostWireEntry.scala
andrew-nowak: Thanks - if you agree with the uncommented version then that's good enough for me - the alternative has been removed

PR #39, opened 2024-09-30
`mediaCatCodes` query param on `/api/search` 

twrichards:
Following a similar pattern/approach to https://github.com/guardian/newswires/pull/30, here we make it possible to search/filter based on `mediaCatCodes` (which are `OR`ed together)

## What's changed

- [convert `mediaCatCodes` to an array at ingestion time](https://github.com/guardian/newswires/commit/a46369882035f2b56c0a663d86b2b6213a59db1a)
- [randomise `mediaCatCodes` for local ingestion](https://github.com/guardian/newswires/commit/c17bbb46bb6c05b8e2aed0c276b1f02d6d562770)
- [expose `mediaCatCodes` from api](https://github.com/guardian/newswires/commit/8de0ae540c275f96bb721032745ca6aa466740cd)
- [add `mediaCatCodes_index` via DB migration](https://github.com/guardian/newswires/commit/1d2994ae9f42ca4c9e1918158c80bf3b211abdf9)
- [add `mediaCatCodes` query param to `/api/search` endpoint and filter results with any of the specified cat codes (i.e. ORed together)](https://github.com/guardian/newswires/commit/35ee19ed750d72b05f26f14b73d5e32861834803)

PR #52, opened 2024-10-10
ability to filter results by supplier

andrew-nowak:
requires categorisation of suppliers by one form or another: currently we can get there purely by source-feed. further suppliers may require more attributes, if eg. multiple small suppliers share a feed

### DIFF
--- a/newswires/app/db/FingerpostWireEntry.scala
+++ b/newswires/app/db/FingerpostWireEntry.scala
@@ -59,25 +59,25 @@ object FingerpostWireEntry extends SQLSyntaxSupport[FingerpostWireEntry] {
   implicit val format: OFormat[FingerpostWireEntry] =
     Json.format[FingerpostWireEntry]
 
-  override val columns =
+  override val columns: Seq[String] =
     Seq("id", "external_id", "ingested_at", "content", "combined_textsearch")
-  val syn = this.syntax("fm")
+  private val syn = this.syntax("wire_entry")
 
   private val selectAllStatement = sqls"""
-    |   ${FingerpostWireEntry.syn.result.id},
-    |   ${FingerpostWireEntry.syn.result.externalId},
-    |   ${FingerpostWireEntry.syn.result.ingestedAt},
-    |   ${FingerpostWireEntry.syn.result.content}
+    |   ${syn.result.id},
+    |   ${syn.result.externalId},
+    |   ${syn.result.ingestedAt},
+    |   ${syn.result.content}
     |""".stripMargin
 
   def apply(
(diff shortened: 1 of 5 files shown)


newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -113,41 +114,48 @@ object FingerpostWireEntry extends SQLSyntaxSupport[FingerpostWireEntry] {
...
+    val whereClause = dataOnlyWhereClauses ++ commonWhereClauses match {
       case Nil        => sqls""
       case whereParts => sqls"WHERE ${sqls.joinWithAnd(whereParts: _*)}"
     }
 
+    val joinClause = if (maybeSupplier.isDefined) {
+      sqls"LEFT OUTER JOIN ${SourceFeedToSupplier as SourceFeedToSupplier.syn} ON ${syn.column("content")}->>'source-feed' = ${SourceFeedToSupplier.syn.sourceFeed}"
+    } else {
+      sqls""
+    }
```
twrichards, line 143:
```suggestion
    val joinClause = maybeSupplier match {
      case Some(_) => sqls"LEFT OUTER JOIN ${SourceFeedToSupplier as SourceFeedToSupplier.syn} ON ${syn.column("content")}->>'source-feed' = ${SourceFeedToSupplier.syn.sourceFeed}"
      case None => sqls""
    }
```

twrichards:
```suggestion
    val joinClause = maybeSupplier.fold(
      sqls""
    )(_ => 
      sqls"LEFT OUTER JOIN ${SourceFeedToSupplier as SourceFeedToSupplier.syn} ON ${syn.column("content")}->>'source-feed' = ${SourceFeedToSupplier.syn.sourceFeed}")
    )
```

twrichards (approved):
code LGTM (made a couple of suggestions)

Is there a reason we must use a DB table to represent this source feed to supplier mapping? can we not represent in Scala code and use the mapping to build just a where clause?? 

Also do we need to add an index to source_feed field of the content column (like was added for keywords and for draft #39)?

andrew-nowak:
We don't _have_ to use a table to drive the relationships - in my head for lots of this wires work is making pages that would allow admins to control definitions of things such as sourcefeed -> supplier without dev intervention, which I think works best if it's backed by a table. Maybe I'm being too premature with that though, and config is good enough for now (I really don't want to hardcode these relationships at any point though)

andrew-nowak:
Spiked getting the mapping from config here: https://github.com/guardian/newswires/pull/60

andrew-nowak:
closing in favour of #60

PR #54, opened 2024-10-14
add button to show JSON

andrew-nowak:
## What does this change?

image

### DIFF
--- a/newswires/client/src/Item.tsx
+++ b/newswires/client/src/Item.tsx
@@ -5,9 +5,11 @@ import {
 	EuiFlyoutFooter,
 	EuiPageTemplate,
 	EuiSpacer,
+	EuiSwitch,
 	EuiTitle,
 	useGeneratedHtmlId,
 } from '@elastic/eui';
+import { css } from '@emotion/react';
 import { useEffect, useState } from 'react';
 import { type WireData, WireDataSchema } from './sharedTypes';
 import { useSearch } from './useSearch';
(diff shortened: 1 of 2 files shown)


newswires/client/src/Item.tsx
```diff
@@ -18,6 +19,7 @@ export const Item = ({ id }: { id: string }) => {
 
 	const [itemData, setItemData] = useState<WireData | undefined>(undefined);
 	const [error, setError] = useState<string | undefined>(undefined);
+	const [showJson, setShowJson] = useState<boolean>(false);
```
twrichards, line 22: pedantic, and you might disagree, but I find `should` or `is` prefixes useful for boolean variables (when I see `showJson` I assume that's a function...
```suggestion
	const [shouldShowJson, setShouldShowJson] = useState<boolean>(false);
```

newswires/client/src/Item.tsx
```diff
@@ -79,10 +81,18 @@ export const Item = ({ id }: { id: string }) => {
...
-					<EuiFlyoutFooter>
+					<EuiFlyoutFooter
+						css={css`
+							display: flex;
+							justify-content: space-between;
+						`}
+					>
 						<EuiButton onClick={() => handleDeselectItem()}>Close</EuiButton>
+						<EuiButton onClick={() => setShowJson(!showJson)}>
+							Show JSON
```
twrichards, line 94: nice to have...
```suggestion
							{shouldShowJson ? "Hide" : "Show"} JSON
```

twrichards: unless EUI has a toggle button, in which case button text could be static

twrichards (approved):
LOVE this!

newswires/client/src/Item.tsx
andrew-nowak: ooh yeah toggle is a good shout! 
image

PR #59, opened 2024-10-15
Richer search UI in sidebar

bryophyta:
## What does this change?

- Add supplier filters and search history to the sidebar.
- Add support on the front end for `subject` searches.
- Updating the text search via the UI happens _within_ the selected set of suppliers, now, rather than overwriting the whole query.

image

## Accessibility

### DIFF
--- a/newswires/client/src/SideNav.tsx
+++ b/newswires/client/src/SideNav.tsx
@@ -50,7 +50,7 @@ export const SideNav = () => {
 			if (activeSuppliers.length === 0) {
 				handleEnterQuery({
 					...config.query,
-					supplier: recognisedSuppliers.filter((s) => s !== supplier),
+					supplier: [supplier],
 				});
 				return;
 			}
--- a/newswires/client/src/SideNav.tsx
+++ b/newswires/client/src/SideNav.tsx
@@ -1,47 +1,93 @@
 import {
+	EuiBadge,
+	EuiBadgeGroup,
 	EuiCollapsibleNav,
 	EuiCollapsibleNavGroup,
 	EuiHeaderSectionItemButton,
 	EuiIcon,
 	EuiListGroup,
+	EuiListGroupItem,
+	EuiText,
 } from '@elastic/eui';
-import { useMemo, useState } from 'react';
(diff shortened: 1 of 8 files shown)


newswires/client/src/SideNav.tsx
```diff
@@ -59,48 +105,129 @@ export const SideNav = () => {
...
+						</EuiBadgeGroup>
+					)}
+				</EuiCollapsibleNavGroup>
 			</EuiCollapsibleNav>
 		</>
 	);
 };
 
-const SearchGroup = ({
+/**
```
bryophyta, line 166: I added the 'singleSearchAsListOfBadges' component in an earlier draft of this branch, where it looked like we were going to add more in-depth support for keyword searches. It seems v likely that we'll add this back in in a future PR, so hopefully this will avoid reinventing the wheel later

newswires/client/src/SideNav.tsx
```diff
@@ -1,47 +1,93 @@
...
-		{ label: 'AAP', query: { q: '', supplier: ['AAP'] } },
-	];
+	const toggleSupplier = useCallback(
+		(supplier: string) => {
+			// If 'activeSuppliers' is empty, that means that *all* suppliers are active.
+			if (activeSuppliers.length === 0) {
+				handleEnterQuery({
+					...config.query,
+					supplier: recognisedSuppliers.filter((s) => s !== supplier),
+				});
```
andrew-nowak, line 54: hmm I think this behaves the opposite to how I would expect - I think if all suppliers are active and I click on AP, I'd expect to filter to just AP, not filter to everything except AP (maybe adding a shift+click or alt+click handler to exclude AP)

andrew-nowak (approved):
looks great! I think the filter behaviour isn't quite natural, but definitely in favour of merging and iterating than holding this branch up

newswires/client/src/SideNav.tsx
paperboyo: Alt would be excluding (as per recent Grid precedent, itself modelled on decades-long Adobe one)

PR #60, opened 2024-10-15
alternative: run the supplier search from a configured mapping of source-feeds to suppliers

andrew-nowak:
## What does this change?

Updates the supplier search method in #52 to use a configured mapping of source-feeds to suppliers, instead of trying to join from a table. Leaves us with simpler db queries (they'll be large for suppliers with many feeds, but easy to generate!) at the expense of not having the potential functionality of letting users configure the mapping via a UI screen. I still think that's a useful feature to work toward, but there's no guarantee we'll get round to it so no need for premature optimisation!

One awkwardness to resolve is that this requires a new config key, but we're frequently running against the CODE database, but I also don't really want to commit this mapping into a public repository like ours... not sure what the best way to square this circle might be. 

One way might be to leave this data in the database, but access it and build the map on launch or schedule, rather than by joining with it at query time (since actually the join seems not-trivial to optimise into an efficient query).


twrichards:
Hey @andrew-nowak - thanks for creating this following my comment on the table based implementation (https://github.com/guardian/newswires/pull/52#pullrequestreview-2368823285). 

Whilst I think this is simpler (and not premature optimisation as you say) compared to https://github.com/guardian/newswires/pull/52 - I had originally thought the mapping was best in the scala code, I see you have concern about that... 

> I also don't really want to commit this mapping into a public repository like ours

... I can't see myself how it would reveal anything sensitive (CC @guardian/newsroom-resilience @blishen @KarenFrankel9) and would have the advantage of PR process to change.

twrichards (approved):
code LGTM, approving to unblock (although expressed some thoughts in https://github.com/guardian/newswires/pull/60#issuecomment-2415256178)

andrew-nowak:
Thanks! Discussed offline and consensus agreed with you, no reason not to have public, so added to Scala code (at least for now!)

PR #61, opened 2024-10-16
Side nav supplier searches

andrew-nowak:
## What does this change?

Updates the Query type, and the URLs that the currently hardcoded side navs link to, to match the new supplier searches added in #60

### DIFF
--- a/newswires/client/src/urlState.ts
+++ b/newswires/client/src/urlState.ts
@@ -1,6 +1,6 @@
 import type { Config, Query } from './sharedTypes';
 
-export const defaultQuery: Query = { q: '' };
+export const defaultQuery: Query = { q: '', supplier: [] };
 
 export const defaultConfig: Config = Object.freeze({
 	view: 'home',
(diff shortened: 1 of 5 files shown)


newswires/client/src/urlState.ts
```diff
@@ -45,14 +47,20 @@ export const configToUrl = (config: Config): string => {
 };
 
 export const paramsToQuerystring = (config: Query): string => {
-	const params = Object.fromEntries(
-		Object.entries(config).reduce<Array<[string, string]>>((acc, [k, v]) => {
+	const params = Object.entries(config).reduce<Array<[string, string]>>(
```
bryophyta, line 50: turning the `Array<[string, string]>>` into an object before creating the `new URLSearchParams` was causing us to remove repeated keys from the params, e.g.

`Object.entries([[a, 1], [b, 2], [a, 3]])` becomes `{b: 2, a: 3}`, rather than `{a: 1, b: 2, c: 3}`.

I can't work out though why this wasn't being caught by the new test case -- `'converts config with many suppliers to querystring'` in this branch...

bryophyta (approved):
looks great to me! as mentioned off-thread I've pushed a couple of small tweaks, but feel free to change them as you see fit 👍

newswires/client/src/urlState.ts
andrew-nowak: turns out it was being caught by the test case :D - I just hadn't run it locally, and the current CI definition doesn't run the client tests (probably due to there not being any before...)
fixed in latest commit

PR #62, opened 2024-10-16
use websearch_to_tsquery for textsearch

andrew-nowak:
Also clearly parse querytext as English

## What does this change?

`phraseto_tsquery` parses the query text with the same idea as googling, or github searching with double quotes around the text - ie. looks for runs of the same words, not just a document which contains each word somewhere, even if there's little or no relation between them. 

`websearch_to_tsquery` parses queries in a similar manner as users familiar with Google would expect - double quotes for phrase searches (as above), `-` for negative searches, and even respects `or` as a keyword for optional searches.


bryophyta (approved):
lgtm!

twrichards (approved):
👏

PR #63, opened 2024-10-16
include newswires client in root package workspace

andrew-nowak:


bryophyta:
re-implemented by #135

PR #64, opened 2024-10-17
include panda-session for automatic session refreshes

andrew-nowak:
## What does this change?

Brings in panda-session so we can get automated session refreshes and keep wires users wiring for longer


twrichards (approved):
nice!

PR #58, opened 2024-10-15
add the snyk workflow

andrew-nowak:

PR #68, opened 2024-10-18
try snyk without setting working directory

andrew-nowak:
Snyk docs say it should autodetect projects within the first few directories, so maybe we don't even need to specify the projects it should be looking at

Attempt to fix up #58


bryophyta (approved):
lgtm, and I can see the snyk action running successfully in other repos [that don't specify a working directory explicitly](https://github.com/guardian/frontend/blob/0042f8d129fa1d7e952ff4144b5442a8127de930/.github/workflows/snyk.yml#L13)

PR #69, opened 2024-10-18
Parameters to exclude wires categories

andrew-nowak:
## What does this change?

As much as you should be able to request to see wires matching suppliers or keywords (and soon subjects), you should _also_ be able to exclude certain suppliers or keywords. 

This is trickier in the DB queries, because postgres seems reluctant or unable to use the indexes that support the positive queries for negative. Thanks to a very handy stack overflow post, there is a trick that allows you to work around this, by querying for the rows which do match the exclusion parameters, then exclude _those_ rows from the real query. Because you're only searching the DB in a positive manner, postgres seems much happier/able to use the indexes for fast querying.

## Accessibility

### DIFF
--- a/newswires/app/controllers/QueryController.scala
+++ b/newswires/app/controllers/QueryController.scala
@@ -1,8 +1,9 @@
 package controllers
 
 import com.gu.pandomainauth.PanDomainAuthSettingsRefresher
+import com.gu.pandomainauth.action.UserRequest
 import com.gu.permissions.PermissionsProvider
-import db.FingerpostWireEntry
+import db.{FingerpostWireEntry, SearchParams}
 import play.api.libs.json.Json
 import play.api.libs.ws.WSClient
 import play.api.mvc.{Action, AnyContent, BaseController, ControllerComponents}
@@ -23,15 +24,24 @@ class QueryController(
       maybeKeywords: Option[String],
       suppliers: List[String],
       maybeBeforeId: Option[Int],
-      maybeSinceId: Option[Int]
-  ): Action[AnyContent] = apiAuthAction {
+      maybeSinceId: Option[Int],
+  ): Action[AnyContent] = apiAuthAction { request: UserRequest[AnyContent] =>
+    val queryParams = SearchParams(
+      text = maybeFreeTextQuery,
+      keywordIncl = maybeKeywords.map(_.split(",").toList).getOrElse(Nil),
+      keywordExcl = request
+        .getQueryString("keywordsExcl")
+        .map(_.split(",").toList)
+        .getOrElse(Nil),
+      suppliersIncl = suppliers,
+      suppliersExcl =
+        request.queryString.get("supplierExcl").map(_.toList).getOrElse(Nil)
+    )
 
     Ok(
       Json.toJson(
         FingerpostWireEntry.query(
-          maybeFreeTextQuery,
-          maybeKeywords.map(_.split(',').toList),
-          suppliers,
+          queryParams,
           maybeBeforeId,
           maybeSinceId,
           pageSize = 30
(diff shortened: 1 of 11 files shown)


newswires/app/controllers/QueryController.scala
```diff
@@ -17,21 +24,34 @@ class QueryController(
...
       maybeBeforeId: Option[Int],
       maybeSinceId: Option[Int]
-  ): Action[AnyContent] = apiAuthAction {
+  ): Action[AnyContent] = apiAuthAction { request: UserRequest[AnyContent] =>
+    val queryParams = SearchParams(
+      text = maybeFreeTextQuery,
+      keywordIncl = maybeKeywords.map(_.split(",").toList).getOrElse(Nil),
+      keywordExcl = paramToList(request, "keywordsExcl"),
+      suppliersIncl = suppliers,
+      suppliersExcl =
```
bryophyta, line 45: is there a reason why we're sometimes using the `paramsToList` method and sometimes not, here?

newswires/client/src/SideNav.tsx
```diff
@@ -18,11 +17,10 @@ import type { Query } from './sharedTypes';
 import { useSearch } from './useSearch';
 
 function decideLabelForQueryBadge(query: Query): string {
-	const { supplier, q, subject } = query;
-	const supplierLabel = supplier.join(', ');
+	const { supplier, q } = query;
+	const supplierLabel = supplier?.join(', ') ?? [];
```
bryophyta, line 21:
```suggestion
	const supplierLabel = supplier?.join(', ') ?? '';
```

newswires/client/src/sharedTypes.ts
```diff
@@ -44,8 +44,12 @@ export type WiresQueryResponse = z.infer<typeof WiresQueryResponseSchema>;
 
 export const QuerySchema = z.object({
 	q: z.string(),
-	supplier: z.array(z.string()),
-	subject: z.array(z.string()),
+	supplier: z.array(z.string()).optional(),
+	supplierExcl: z.array(z.string()).optional(),
+	keywords: z.ostring(),
```
bryophyta, line 49: I wonder whether we'd want to have these as arrays of strings too, further down the line, in order to make it easier to handle them in the UI? But not necessary at this stage afaict

newswires/client/src/urlState.test.ts
```diff
@@ -56,6 +60,75 @@ describe('urlToConfig', () => {
 			query: { ...defaultQuery, q: 'abc', supplier: [] },
 		});
 	});
+
+	it('can exclude multiple suppliers', () => {
+		const url = makeFakeLocation('/feed?q=abc&supplierExcl=PA');
```
bryophyta, line 65: there's only one supplier being excluded here

bryophyta (approved):
A couple of small points but this looks good to me!

Great work discovering how to get the exclusions to work with the indices!

newswires/client/src/sharedTypes.ts
andrew-nowak: yep - broadly I think depends whether we want to standardise on `&keyword=a&keyword=b&keyword=c` or `&keywords=a,b,c` - both fine but we should pick one and not do both... later

newswires/app/controllers/QueryController.scala
andrew-nowak: they're subtly different, bun incompatible:
currently `supplier` and `supplierExcl` are repeated query params, rather than single params with comma-separated values. the former is extracted by `request.queryString.get(X)`, the latter by `request.getQueryString(X)` (which is what is in `paramToList`)

keywords is extracted by the mapping in the routes file, rather than directly from the request object here... we should also standardise on the location for that - putting them in the routes file is nice because it can then serve as documentation not only of routes but available parameters, but I'm a little nervous because it's a lot of params of the same type, so it'd be easy to accidentally bind a parameter to the wrong name, which would be painful to debug, so I'm putting that decision off too 😅

PR #70, opened 2024-10-20
If we're refreshing data, pass 'sinceId'

bryophyta:
Make use of the API 👍

### DIFF
--- a/newswires/client/src/useSearch.tsx
+++ b/newswires/client/src/useSearch.tsx
@@ -186,8 +186,14 @@ function reducer(state: State, action: Action): State {
 	}
 }
 
-async function fetchResults(query: Query): Promise<WiresQueryResponse> {
-	const queryString = paramsToQuerystring(query);
+async function fetchResults(
+	query: Query,
+	sinceId: string | undefined = undefined,
+): Promise<WiresQueryResponse> {
+	const queryToSerialise = sinceId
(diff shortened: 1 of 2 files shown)


newswires/client/src/useSearch.tsx
```diff
@@ -291,7 +297,12 @@ export function SearchContextProvider({ children }: PropsWithChildren) {
 		if (state.status === 'success') {
 			pollingInterval = setInterval(() => {
 				// Poll for updated results
-				fetchResults(currentConfig.query)
+				fetchResults(
+					currentConfig.query,
+					Math.max(
+						...state.queryData.results.map((wire) => wire.id),
```
andrew-nowak, line 303: let's not rely on the id being a monotonically increasing value - instead let's find the max by the attribute we actually want to sort on (probably one of the dates) - or just choose whichever content is in the 0th position

bryophyta: I'm happy with that in principle, but I feel like it makes sense given that this is the way the back-end API [currently works](https://github.com/guardian/newswires/blob/195ea1f3edd9f9f4e85398ec8938f2c7d1715b4d/newswires/app/db/FingerpostWireEntry.scala#L149-L151)?

andrew-nowak: gah, great point - let's address this later

andrew-nowak (approved):

PR #71, opened 2024-10-21
Put search in side nav

bryophyta:
## What does this change?

- Move search bar to side nav
- Refactor `SearchBox` component so that it consumes global values from `useSearch` directly, rather than having them passed as children.

## Images

image

## Accessibility

### DIFF
--- a/newswires/client/src/SideNav.tsx
+++ b/newswires/client/src/SideNav.tsx
@@ -107,7 +107,7 @@ export const SideNav = () => {
 				}
 				onClose={() => setNavIsOpen(false)}
 			>
-				<SearchBox incremental={true} />
+				<SearchBox />
 				<EuiCollapsibleNavGroup title={'Suppliers'}>
 					<EuiListGroup
 						maxWidth="none"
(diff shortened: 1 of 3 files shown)


newswires/client/src/SideNav.tsx
```diff
@@ -105,6 +107,7 @@ export const SideNav = () => {
 				}
 				onClose={() => setNavIsOpen(false)}
 			>
+				<SearchBox incremental={true} />
```
andrew-nowak, line 110: we only use the SearchBox here I think - do we need an `incremental` flag?

andrew-nowak (approved):

newswires/client/src/SideNav.tsx
bryophyta: good call, it doesn't seem likely that we'll reintroduce the non-incremental version any time soon so have removed the option here: https://github.com/guardian/newswires/pull/71/commits/dd688c43f2bf31039cbf2718825a68c0a160870f

PR #72, opened 2024-10-22
add bucket searches

andrew-nowak:
## What does this change?

WIP: adding "buckets" of searches

I imagine that one day these would live in a database with a short form UI allowing edits without code changes...
For now leave them as hardcoded


newswires/app/conf/SearchBuckets.scala
```diff
@@ -0,0 +1,73 @@
...
+    "iptccat:SDP",
+    "iptccat:SDR",
+    "iptccat:SSS",
+    "iptccat:SRS",
+    "iptccat:SRZ",
+    "iptccat:SDS",
+    "iptccat:SDQ",
+    "iptccat:SCR",
+    "iptccat:SST",
+    "iptccat:RSR",
```
bryophyta, line 23: I suspect that `iptccat:RDR` and `iptccat:RRR` are for horse racing (I did a couple of searches and that's all it returned) but not sure how to confirm their meaning..?

bryophyta (approved):
this is so great1 I've tested locally and it all looks good to me 👍

PR #75, opened 2024-10-22
Highlight search terms in results

bryophyta:
## What does this change?

- Add a `highlight` column, that should return a `ts_headline` for the given query, if there's a free text search as part of that query
- Display the highlight column in the main feed table, and use the highlighted `bodyText` as the markup for the main item display, if we have it.
  - nb. the highlighting is achieved by adding `<mark>` elements either side of the search terms, which means that the markup in the body text on the page won't be quite the same as the markup that we have stored in the db for the story. We should bear this in mind when working on the 'import to composer' via pinboard functionality (cc. @twrichards) though I don't think it's a huge issue. 
- Tweak the main feed table to use a 'responsive' layout whenever an item is selected, as there's reduced horizontal space but the standard responsive layout doesn't kick in because the overall page is still at a wide breakpoint.

## Images

image

## Accessibility

### DIFF
--- a/newswires/app/db/FingerpostWireEntry.scala
+++ b/newswires/app/db/FingerpostWireEntry.scala
@@ -65,7 +65,8 @@ case class FingerpostWireEntry(
     id: Long,
     externalId: String,
     ingestedAt: ZonedDateTime,
-    content: FingerpostWire
+    content: FingerpostWire,
+    highlight: Option[String] = None
 )
 
 object FingerpostWireEntry
@@ -76,7 +77,14 @@ object FingerpostWireEntry
     Json.format[FingerpostWireEntry]
 
   override val columns =
-    Seq("id", "external_id", "ingested_at", "content", "combined_textsearch")
+    Seq(
+      "id",
+      "external_id",
+      "ingested_at",
+      "content",
+      "combined_textsearch",
+      "highlight"
+    )
   val syn = this.syntax("fm")
 
   private val selectAllStatement = sqls"""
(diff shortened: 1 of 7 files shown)


newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -151,24 +172,35 @@ object FingerpostWireEntry extends SQLSyntaxSupport[FingerpostWireEntry] {
       )
     ).flatten
 
+    val highlightsClause = maybeFreeTextQuery match {
+      case Some(query) =>
+        sqls", ts_headline('simple', $contentCol->>'body_text', phraseto_tsquery($query)) AS ${syn.result.highlight}"
+      case None => sqls""
```
andrew-nowak, line 178: this seems to be really awkward; the fixes might be to run a completely different query when we expect a highlight back (and extracting all the common query bits into something shared), but we can also go hack with
```suggestion
      case None => sqls"'' AS ${syn.resultName.highlight}"
```
(ie. always return an empty string as a "highlight", even if there is no query to highlight)

bryophyta: ooh, interesting! thanks for looking at this!

newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -78,15 +86,26 @@ object FingerpostWireEntry extends SQLSyntaxSupport[FingerpostWireEntry] {
...
 
-  def get(id: Int): Option[FingerpostWireEntry] = DB readOnly {
+  def get(
+      id: Int,
+      maybeFreeTextQuery: Option[String] = None
+  ): Option[FingerpostWireEntry] = DB readOnly {
+    val contentCol = FingerpostWireEntry.syn.column("content")
+    val highlightsClause = maybeFreeTextQuery match {
+      case Some(query) =>
+        sqls""", ts_headline('simple', $contentCol->>'body_text', websearch_to_tsquery($query), 'StartSel=<mark>, StopSel=</mark>, HighlightAll=true') AS h_on_fm"""
```
andrew-nowak, line 103: searching on 'simple' will return only matches that match exactly - if there is something that has a matching stem then the highlighter won't find anything to return and the results look odd; try searching for 'Londoner' - you'll get lots of results for 'London', but because it doesn't match the highlight is just for the first few words in the body

This might be the desired behaviour, in which case we should not display the highlight; but I think it'd be best if we used the 'english' "dictionary" -- even if the results aren't exactly what the user entered, they can at least see what we thought was similar enough to return this wire (and can complain if it's nonsensical and we can try to improve)

```suggestion
        sqls""", ts_headline('english', $contentCol->>'body_text', websearch_to_tsquery('english', $query), 'StartSel=<mark>, StopSel=</mark>, HighlightAll=true') AS h_on_fm"""
```

bryophyta: ah, fantastic, thanks! I had a todo to try and work out why it was sometimes returning just the first few lines, but hadn't got round to it yet!

andrew-nowak (approved):
lgtm

PR #80, opened 2024-10-24
Add more sports codes

andrew-nowak:
some sports stories are still sneaking into "no sports"... add more newly seen codes


bryophyta (approved):
🥳

PR #82, opened 2024-10-28
Add two racing codes to no sports bucket

bryophyta:
I haven't been able to find documentation of these codes, but they look like they only carry horse racing content.


andrew-nowak:
these and more are in https://github.com/guardian/newswires/pull/80

bryophyta:
> oh wait, is ipccat being included in the sent wires? not iptccat?

Nope you're right! typo on my side. Closing in favour of #80

PR #83, opened 2024-10-28
WIP: `poller-lambdas`

twrichards:
## Accessibility


poller-lambdas/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+		}
+		return Promise.all(
+			Records.map((record) => {
+				const valueFromPreviousPoll = record.body; //TODO maybe parse
+				return pollerFunction(secret, valueFromPreviousPoll)
+					.then(async (output) => {
+						const endTimeEpochMillis = Date.now();
+
+						await sqs.send(
+							new SendMessageCommand({
```
bryophyta, line 46: this needs to accommodate adding multiple items to the queue, from one poller lambda payload

poller-lambdas/src/types.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+export type LongPollOutput = CorePollerOutput & { valueForNextPoll: string };
+
+export type FixedFrequencyPollOutput = CorePollerOutput & {
+	idealFrequencyInSeconds: number;
+};
+
+export type LongPollFunction = (
+	secret: SecretValue,
+	input: PollerInput,
```
bryophyta, line 16:
```suggestion
	input: PollerInput | undefined,
```

my current understanding is that input needs to be potentially undefined, to account for the case where a new deployment needs to bootstrap itself (i.e. where there's no pre-existing poller output to take in).

given that the SQS invocation needs to set some body value every time (as per https://github.com/guardian/newswires/pull/83/files#diff-fde74c4679830198a8758c56ef27dc2a92722727d787a4d36a93f82068f80af6R71) it looks like the poller-lambda wrapper will need to do something to distinguish between the case where there's a valid PollerInput and where there isn't

bryophyta: We discussed this off-thread; because we want to only have one 'chain' of requests for a given poller at any given time (to avoid duplication of work). So bootstrapping a new chain of tasks in the SQS queue that drives the lambda should be done intentionally and therefore it may make sense to just pass the 'last known good value' manually when we start it. (This isn't a super clear description; I need to log off now but will try to revisit to clarify tomorrow.)

poller-lambdas/src/aws.ts
```diff
@@ -0,0 +1,44 @@
...
+	POLLER_LAMBDA_ENV_VAR_KEYS.OWN_QUEUE_URL,
+);
+
+export const queueNextInvocation = (props: {
+	MessageBody: string;
+	DelaySeconds?: number;
+}) =>
+	sqs.send(
+		new SendMessageCommand({
+			QueueUrl: ownQueueUrl,
```
bryophyta, line 40: when testing locally, `ownQueueUrl` here is false, and I believe that this is because it's set when this file is imported by `localRun`, which is before the env var has been set, within the main run function of `localRun`.

if we move line 30 to inside the `queueNextInvocation` block then it will get the right value when run locally

poller-lambdas/localRun.ts
```diff
@@ -0,0 +1,131 @@
...
+		type: 'text',
+		name: 'input',
+		message:
+			"Do want to provide a starting input? (e.g. if this is poller is 'long polling')",
+	})) as { input: string };
+
+	await fakeInvoke(
+		handler,
+		input,
+		sqs.queueData[ownQueueUrl]!,
```
bryophyta, line 128: when this line is run, afaict, there are no keys in `queueData`, so `sqs.queueData[ownQueueUrl]!,` evaluates to `undefined` and stays that way..

could potentially either pre-populate the queue with an empty array, or else do the `queueData` lookup at the time it needs to be used in the body of `fakeInvoke`?

poller-lambdas/src/index.ts
twrichards: done

poller-lambdas/src/aws.ts
twrichards: many thanks Pete, I've addressed this amongst other things in https://github.com/guardian/newswires/pull/83/commits/841e6f0fc31dcf920a7e416a9e33e5219464afe0

poller-lambdas/localRun.ts
twrichards: many thanks Pete, I've addressed this amongst other things in https://github.com/guardian/newswires/pull/83/commits/841e6f0fc31dcf920a7e416a9e33e5219464afe0

cdk/bin/cdk.ts
```diff
@@ -1,47 +1,65 @@
...
 }).addDependency(codeWiresFeeds);
 
 new Newswires(app, 'Newswires-PROD', {
 	env,
-	stack,
+	stack: STACK,
 	stage: 'PROD',
 	domainName: 'newswires.gutools.co.uk',
 	enableMonitoring: false,
+	sourceQueue: prodWiresFeeds.sourceQueue,
```
bryophyta, line 47: is there a particular reason for creating the source queue as part of a different stack?

cdk/cdk.json
```diff
@@ -1,5 +1,5 @@
 {
-	"app": "tsx bin/cdk.ts",
+	"app": "ts-node-dev bin/cdk.ts",
```
bryophyta, line 2: we discussed off-thread that the reason for using `ts-node-dev` here is because `tsx` restarts if the user presses Enter, which doesn't work with the current CLI setup

docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
   localstack:
     image: localstack/localstack:latest
     environment:
-      - SERVICES=s3
+      - SERVICES=s3,sqs
```
bryophyta, line 7: are we using localstack sqs in the end?

poller-lambdas/README.md
```diff
@@ -0,0 +1,18 @@
+This directory/module contains the logic for each of out pollers and an abstraction for running them. Configured pollers also have their infrastructure generated automatically via CDK 🎉
+
+Poller lambdas have their own SQS queues to facilitate long-polling and to allow invoking on a frequency faster than the once per minute allowed by CloudWatch/EventBridge rules.
+
+## Adding a new poller lambda
```
bryophyta, line 5: 💜🙌

cdk/bin/cdk.ts
twrichards: I think because FIP have a reference to it, so I think we wanted to decouple, such that we could theoretically tear down the newswires stack without impacting FIP writing to the queue. @andrew-nowak could probably confirm...

docker-compose.yml
twrichards: ahh, no we're not, good spot, thanks, I've removed that in https://github.com/guardian/newswires/pull/83/commits/1df356325c7249eeb2613e647f2b6f4c823bc394

cdk/lib/constructs/pollerLambda.ts
```diff
@@ -0,0 +1,176 @@
...
+			pollerConfig.idealFrequencyInSeconds > Duration.minutes(15).toSeconds()
+		) {
+			throw Error(
+				"SQS' max delay is 15mins, so if you really mean to poll so infrequently (e.g. hourly or daily), then you'll need to use Cloudwatch/EventBridge rules instead of SQS ",
+			);
+		}
+
+		// only one lambda should be happening at once, however the lambda queues up its next execution whilst it's still running
+		// so concurrency of 2 should prevent throttling, but also guard against it going haywire
+		const reservedConcurrentExecutions = 2;
```
andrew-nowak, line 69: I don't follow what's wrong with setting reserved concurrency to 1 - yes the throttling metric will be non-zero but if that's expected behaviour then we can simply ignore the metric, and use a more useful metric to detect poor performance (ie. queue depth or age of oldest queue item, etc.)

poller-lambdas/src/index.ts
```diff
@@ -0,0 +1,110 @@
...
+				})
+				.catch((error) => {
+					console.error('FAILED', error);
+					// consider still queuing next (perhaps with default delay or 1min) to avoid the lambda from stopping entirely
+					throw error;
+				});
+		}
+	};
+
+export = {
```
andrew-nowak, line 104: shouldn't this be `export {` ? I didn't think this was valid for esmodules - confused why the compiler isn't complaining

poller-lambdas/src/config.ts
```diff
@@ -0,0 +1,23 @@
...
+			region: 'eu-west-1',
+			credentials: fromNodeProviderChain({
+				profile: 'editorial-feeds',
+			}),
+		}
+	: {};
+
+export const getEnvironmentVariableOrCrash = (
+	key: keyof typeof POLLER_LAMBDA_ENV_VAR_KEYS,
+) => process.env[key]!;
```
andrew-nowak, line 23: kind of a nit: this won't crash if the var is unset, but it'll assert that it was set, and the caller may (or may not!) crash when it tries to use the var. it might be nicer to forcefully throw from here instead? which might make following the stack trace easier to follow

poller-lambdas/src/aws.ts
```diff
@@ -0,0 +1,43 @@
...
+export const queueNextInvocation = (props: {
+	MessageBody: string;
+	DelaySeconds?: number;
+}) =>
+	sqs.send(
+		new SendMessageCommand({
+			QueueUrl: getEnvironmentVariableOrCrash(
+				POLLER_LAMBDA_ENV_VAR_KEYS.OWN_QUEUE_URL,
+			),
+			MessageDeduplicationId: lambdaApp, // should prevent the same lambda from being invoked multiple times
```
andrew-nowak, line 40: do we still want this if the queue is no longer fifo?

cdk/lib/constructs/pollerLambda.ts
twrichards: it's a fair question. I think the queue metrics are a bit laggy and the also the GuLambda construct gives us an easy alarm for throttling. I don't love the idea that it would always be throttling, as I think it might attempt retries on throttle. on balance I think this limit of two is the cleanest

poller-lambdas/src/config.ts
twrichards: wow! TIL that the 'bang' doesn't actually verify at runtime, but rather its cue from the developer to compile to assume it's not nullish 🫨 (I use it quite often and I need to stop). I've replaced it in https://github.com/guardian/newswires/pull/83/commits/f914a118e0a9250098982eedc1997491be191708

poller-lambdas/src/aws.ts
twrichards: good spot, thanks, removed in https://github.com/guardian/newswires/pull/83/commits/910b766ed25cafee884e7760732392b7fe52c1b3

PR #84, opened 2024-11-05
Don't write items with no external id to db

bryophyta:
## What does this change?

When the ingestion lambda receives an item from the SQS queue, don't try to write it to the db unless there's the `Message-Id` attribute on the item, to use as an `external_id`. 

Instead:

- Write the item data to S3
- Throw an error

The error will be caught, meaning that the record will be failed, and the reason will be logged. The record will get retried and will almost certainly fail again, but after a couple of tries it'll go into the DLQ.

We should probably tighten up the error handling at some point so that we don't needlessly retry these records, but this feels like a step in the right direction at the moment, and I suggest that we revisit this when we add more validation to the incoming items, within the ingestion lambda.

## How to test

- [x] Deploy to CODE, check that we're still ingesting data as normal.

## Accessibility

### DIFF
--- a/ingestion-lambda/src/handler.ts
+++ b/ingestion-lambda/src/handler.ts
@@ -78,8 +78,19 @@ export const main = async (event: SQSEvent): Promise<SQSBatchResponse> => {
 							messageAttributes['Message-Id']?.stringValue;
 
 						if (!fingerpostMessageId) {
-							console.warn(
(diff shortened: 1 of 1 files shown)


twrichards (approved):
nice. I'm going to merge as I'd like this functionality please 🙏

PR #85, opened 2024-11-05
add riff-raff dependency for Newswires cfn deploy on the WiresFeeds cfn deploy (plus snapshot test the riff-raff.yaml)

twrichards:
The Newswires stack references the queue exposed by the WiresFeeds stack, cdk automagically does this via cfn outputs under the hood. The generated riff-raff.yaml has no idea about this relationship but the two stacks have separate cfn deployments, which by default execute in parallel, creating a race condition.

This PR...
- first snapshot tests the generated riff-raff yaml (as that's useful in its own right)
- second, it makes use of `addDependency` (added in https://github.com/guardian/cdk/pull/2094) to add that dependency (which shows up in the test snapshot)

### DIFF
--- a/cdk/lib/riffraff.test.ts
+++ b/cdk/lib/riffraff.test.ts
@@ -1,0 +1,10 @@
+import { riffraff } from '../bin/cdk';
+
+describe('generated riff-raff.yaml', () => {
+	it('matches the snapshot', () => {
+		// @ts-ignore
+		const outdir = riffraff.outdir; // this changes for every test execution and best not to change cdk.ts too much
+		const riffRaffYaml = riffraff.toYAML().replaceAll(outdir, 'cdk.out');
+		expect(riffRaffYaml).toMatchSnapshot();
+	});
+});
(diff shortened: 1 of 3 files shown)


cdk/lib/riffraff.test.ts
```diff
@@ -0,0 +1,10 @@
+import { riffraff } from '../bin/cdk';
+
+describe('generated riff-raff.yaml', () => {
+	it('matches the snapshot', () => {
+		// @ts-ignore
+		const outdir = riffraff.outdir; // this changes for every test execution and best not to change cdk.ts too much
```
akash1810, line 6: minor: this property is set from the `outdir` of the parent `App`. If we were to instantiate an `App` in the test, we could set the `outdir` prop and thus become deterministic. For example https://github.com/guardian/cdk/blob/ec0fb851e52d6c2b966b6b264eb21dc6a6186cfe/src/riff-raff-yaml-file/index.test.ts#L16.

akash1810 (approved):
Minor comment on the test, but LGTM!

Maybe take my approval with a pinch of salt, as I'm not in the team and thus might be missing some context.

cdk/lib/riffraff.test.ts
twrichards: I prefer the idea of snapshot testing the actual stacks we create, rather than some arbitrarily configured test one.

PR #86, opened 2024-11-12
Initial AP poller implementation

bryophyta:
## What does this change?

Adds a new poller lambda, using the 'long polling' pattern added in #83, to get data from the AP `/feed` endpoint and pass it to the Newswires ingestion lambda. Getting the data needed for the ingestion lambda requires two steps:

1. Long-polling request to the `/feed` endpoint, which returns immediately if there are new items since the last request (determined by the `next_page` value in the response from the previous request, and passed forward to the lambda via the SQS queue that invokes it) or times out after 15 seconds. ([See AP docs](https://api.ap.org/media/v/docs/Feed.htm))
2. The `/feed` endpoint returns only part of the data we're interested in. To get the 'body' of the story we need to fetch and parse the [NITF representation of the story](https://www.iptc.org/std/NITF/3.4/documentation/nitf-documentation.html), which is in an XML format.

**nb.** we don't yet have a clear standard for judging when our processing of a feed is complete/production-ready, so it's possible that we'll want to make some changes later (e.g. if we discover we're missing important fields).

## How to test

- [x] Run locally, using `npm run dev -w poller-lambdas` and selecting the AP poller.
- [x] Deploy to CODE; check that it queues up the next invocation with `next_page` as expected.
- [x] Run the unit tests which check that we're extracting the expected values from some samples of the NITF rendition of a couple of stories.

## Accessibility


bryophyta:
Closing in favour of splitting up into #93, #97 and #98

PR #92, opened 2024-12-19
Reuters poller business logic

bryophyta:
## What does this change?

First pass at business logic for Reuters feed.

#### 1. Auth

  * Read auth token from Secrets
  * If the auth token fails, try to request a new one to use for subsequent calls using the client id and secret from Secrets

#### 2. Request feed via search for the last `n*1.2` seconds, where `n` is the frequency at which the poller runs.

  * nb. as per our conversation with Reuters' API team there is not currently a supported method for getting 'only new items since last poll', so this is my suggested heuristic for getting new stories, though it's not foolproof
  * this means there will be overlap between any two consecutive queries, but currently opting to rely on the ingestion lambda to deduplicate stories unless this proves problematic for performance or other reasons

#### 3. Pass the most recent auth token back

The poller lambda wrapper, as of #99, will check this value and update the Secret value if needed.

---

The implementation of (1) adds a little complexity and mutation that it would be nice to avoid, but the API guidelines suggest we shouldn't request new access tokens too often, so we should respect that. It could be refactored so that e.g. `fetchWithReauth` and `fetchAllPages` were defined outside the scope of the main function, and had `accessToken` passed as an argument, but on reflection it felt to me as though declaring `accessToken` via `let` makes it more explicit a) that it can mutate, and b) which functions can mutate it. I'm open to other perspectives though.

### Script

Also adds a script to help get raw item data from the Reuters API, to help with debugging.

## How to test

- [x] Run locally
- [x] Deploy to CODE, see that Reuters stories are being processed as expected.

## Accessibility

### DIFF
--- a/poller-lambdas/src/pollers/reuters/auth.ts
+++ b/poller-lambdas/src/pollers/reuters/auth.ts
@@ -1,0 +1,39 @@
+import { z } from 'zod';
+
+const AuthSchema = z.object({
+	access_token: z.string(),
+	expires_in: z.number(),
+	token_type: z.string(),
+});
+
+const scopes =
+	'https://api.thomsonreuters.com/auth/reutersconnect.contentapi.read https://api.thomsonreuters.com/auth/reutersconnect.contentapi.write';
+const authUrl = 'https://auth.thomsonreuters.com/oauth/token';
+const grantType = 'client_credentials';
+const audience = '7a14b6a2-73b8-4ab2-a610-80fb9f40f769';
+
+export async function auth(
+	clientId: string,
+	clientSecret: string,
+): Promise<string> {
+	const req = new Request(authUrl, {
+		method: 'POST',
+		headers: {
+			'content-type': 'application/x-www-form-urlencoded',
+		},
+		body: `grant_type=${grantType}&client_id=${clientId}&client_secret=${clientSecret}&audience=${audience}&scope=${encodeURIComponent(scopes)}`,
+	});
+	try {
+		console.log('Requesting new auth token from Reuters');
+		const response = await fetch(req);
+		const data = (await response.json()) as unknown;
+		const { access_token, expires_in } = AuthSchema.parse(data);
+		console.log(
+			`Received new auth token from Reuters, expires in ${expires_in} seconds`,
+		);
+		return access_token;
+	} catch (error) {
+		console.error(error);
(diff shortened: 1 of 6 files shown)


poller-lambdas/src/pollers/reuters/reutersPoller.ts
```diff
@@ -1,18 +1,299 @@
...
-	await new Promise((resolve) => setTimeout(resolve, 1000));
+	const parsedSecret = SecretValueSchema.safeParse(JSON.parse(secret));
+	if (!parsedSecret.success) {
+		throw new Error('Failed to parse secret value for Reuters poller');
+	}
+	const { CLIENT_ID, CLIENT_SECRET, ACCESS_TOKEN } = parsedSecret.data;
+
+	let accessToken = ACCESS_TOKEN ?? (await auth(CLIENT_ID, CLIENT_SECRET));
+
+	async function fetchWithReauth(query: string) {
```
bryophyta, line 201: this function is defined within the scope of the main poller function so that it can read and mutate `accessToken`, I went back and forth on whether to handle this in a more 'pure' way, but as mentioned in the PR description this felt like a good balance in context. Open to other perspectives though!

poller-lambdas/src/pollers/reuters/auth.ts
```diff
@@ -0,0 +1,39 @@
...
+	access_token: z.string(),
+	expires_in: z.number(),
+	token_type: z.string(),
+});
+
+const scopes =
+	'https://api.thomsonreuters.com/auth/reutersconnect.contentapi.read https://api.thomsonreuters.com/auth/reutersconnect.contentapi.write';
+const authUrl = 'https://auth.thomsonreuters.com/oauth/token';
+const grantType = 'client_credentials';
+const audience = '7a14b6a2-73b8-4ab2-a610-80fb9f40f769';
```
andrew-nowak, line 13: just double checking that this value is fine to be made public? iirc it's a value that's shared by all customers, is that right?

andrew-nowak (approved):
lgtm! great work!

poller-lambdas/src/pollers/reuters/auth.ts
bryophyta: yep that's right -- it's from the Reuters API documentation 👍

PR #98, opened 2025-01-13
AP poller business logic

bryophyta:
## What does this change?

Adds business logic for the AP poller, split out from the original PR #86, to get data from the AP `/feed` endpoint and pass it to the Newswires ingestion lambda. Getting the data needed for the ingestion lambda requires two steps:

1. Long-polling request to the `/feed` endpoint, which returns immediately if there are new items since the last request (determined by the `next_page` value in the response from the previous request, and passed forward to the lambda via the SQS queue that invokes it) or times out after 15 seconds. ([See AP docs](https://api.ap.org/media/v/docs/Feed.htm))
2. The `/feed` endpoint returns only part of the data we're interested in. To get the 'body' of the story we need to fetch and parse the [NITF representation of the story](https://www.iptc.org/std/NITF/3.4/documentation/nitf-documentation.html), which is in an XML format.

**nb.** we don't yet have a clear standard for judging when our processing of a feed is complete/production-ready, so it's possible that we'll want to make some changes later (e.g. if we discover we're missing important fields).

## Accessibility


poller-lambdas/src/pollers/ap/parseNitfContent.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+export function nitfBlockToHtml(block: HTMLElement): string {
+	const clonedBlock = block.clone() as HTMLElement;
+	Array.from(clonedBlock.querySelectorAll('hl2')).forEach(
+		(hl2) => (hl2.tagName = 'h2'),
+	);
+	Array.from(clonedBlock.querySelectorAll('media,media-reference')).forEach(
+		(mediaTag) => mediaTag.remove(),
+	);
```
andrew-nowak, line 20: It looks like node-html-parser already returns an array from `querySelectorAll`, so no need to construct another for the loop

```suggestion
	clonedBlock.querySelectorAll('hl2').forEach(
		(hl2) => (hl2.tagName = 'h2'),
	);
	clonedBlock.querySelectorAll('media,media-reference').forEach(
		(mediaTag) => mediaTag.remove(),
	);
```

but personally I find plain for-loops easier to follow for this kind of mutation

```suggestion
	for (const hl2 of clonedBlock.querySelectorAll('hl2')) {
		hl2.tagName = 'h2';
	}
	for (const mediaTag of clonedBlock.querySelectorAll('media,media-reference')) {
		mediaTag.remove();
	}
```

bryophyta: oh nice! I'd assumed it returned a `NodeList` like the DOM method does 👍

poller-lambdas/src/pollers/ap/apPoller.ts
```diff
@@ -1,10 +1,197 @@
...
+
+type FeedItemWithContent = {
+	feedItem: Contentitem;
+	originalXmlContent: string;
+	contentFromNitf: ContentFromNitf;
+};
+
+// https://api.ap.org/media/v-preview/content/feed?page_size=10&in_my_plan=true
+export const apPoller = (async (secret: SecretValue, input: PollerInput) => {
+	// todo: remove '-preview' from baseUrl when we go live
```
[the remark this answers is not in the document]
bryophyta: yep that makes sense, I've done that now 👍

andrew-nowak (approved):

PR #105, opened 2025-01-20
AP poller: link AP images to wire stories

sb-dev:
## What does this change?

Link AP images to wire stories.

## Accessibility

### DIFF
--- a/poller-lambdas/src/pollers/ap/apPoller.ts
+++ b/poller-lambdas/src/pollers/ap/apPoller.ts
@@ -169,6 +169,7 @@ function itemWithContentToDesiredOutput({
 		ednote,
 		subject,
 		keywords,
+		associations,
 	} = feedItem;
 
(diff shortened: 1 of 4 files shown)


poller-lambdas/src/pollers/ap/apPoller.ts
```diff
@@ -191,6 +192,11 @@ function itemWithContentToDesiredOutput({
 			abstract,
 			originalContentText: originalXmlContent,
 			ednote,
+			associations: associations
```
bryophyta, line 195: I'd be tempted to call this field `imageIds`, because at the moment we don't have a more general schema for representing associations so I think it makes sense to only store image ids in this field. What do you think?

Assuming we do that, we'd also want to filter this list for `.type === 'picture'`

andrew-nowak (approved):

PR #108, opened 2025-01-23
Newswires UI: add pagination

sb-dev:
## What does this change?

Add a load more button to view older stories.

## Accessibility

### DIFF
--- a/newswires/client/src/WireItemTable.tsx
+++ b/newswires/client/src/WireItemTable.tsx
@@ -14,6 +14,7 @@ import {
 	useEuiTheme,
 } from '@elastic/eui';
 import { css } from '@emotion/react';
+import { useState } from 'react';
 import sanitizeHtml from 'sanitize-html';
 import { useSearch } from './context/SearchContext.tsx';
 import { formatTimestamp } from './formatTimestamp';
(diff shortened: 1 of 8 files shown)


newswires/client/src/context/SearchContext.tsx
```diff
@@ -167,15 +180,34 @@ export function SearchContextProvider({ children }: PropsWithChildren) {
 				});
 		}
 
-		if (state.status === 'success' || state.status === 'offline') {
+		if (state.status === 'loading-more') {
+			fetchResults(currentConfig.query, {
```
bryophyta, line 184: I wonder whether it would be more straightforward to put the `fetch().then().catch()` into the body of the `loadMoreResults` function? It will hopefully keep the number of branches in the `status` state machine simpler, and as far as I can tell it's only the load more button itself that needs to know about whether the fetch is underway, so we could perhaps have 'loading-more' as a piece of more local state. What do you think?

newswires/client/src/WireItemTable.tsx
```diff
@@ -31,38 +33,63 @@ const fadeOutBackground = css`
...
 
 	const selectedWireId = config.itemId;
 
+	const handleLoadMoreResults = () => {
+		if (wires.length > 0) {
+			setIsLoading(true);
+
+			const beforeId = Math.min(...wires.map((wire) => wire.id)).toString();
+
+			void loadMoreResults(beforeId).finally(() => {
```
bryophyta, line 48: this seems like a great way of handling the async state update 🤩

one small suggestion is to maybe rename `isLoading` to `isLoadingMore`, just to fit with the `loadMoreResults` naming and differentiate it from the global `loading` status? But this isn't a blocking suggestion

newswires/client/src/context/SearchReducer.test.ts
```diff
@@ -98,6 +98,31 @@ describe('SearchReducer', () => {
 		});
 	});
 
+	it(`should handle APPEND_RESULTS action in loading-more state`, () => {
```
bryophyta, line 101:
```suggestion
	it(`should handle APPEND_RESULTS action in success state`, () => {
```

newswires/client/src/WireItemTable.tsx
sb-dev: I agree, I nearly used that name :)

PR #110, opened 2025-01-24
Use anchor tags for feed

bryophyta:
## What does this change?

The main change is to refactor the feed view so that it uses a list of links (`<a>` tags) instead of a table:

* Clicking on an item is more like navigation than on-page action, as it takes you to a view corresponding to a different URL
* Allows us to leverage the platform for `cmd+click` behaviour etc.
* Also allows for more control over layout of the cards, without having to wrangle too much with Eui. In particular, it makes it easier to have a more compact layout.

Additional changes:

* Add scroll-into-view behaviour (again, easier when we're not competing with Eui's features for this, I think)
* Adapt the card heading to use an `<EuiBadge>` to display supplier info, rather than including it at the end of the headline/slug.

## Images

Before and after

image

## Accessibility


bryophyta:
re-implements some of #81

newswires/client/src/context/SearchContext.tsx
```diff
@@ -273,6 +280,32 @@ export function SearchContextProvider({ children }: PropsWithChildren) {
 		dispatch({ type: 'TOGGLE_AUTO_UPDATE' });
 	};
 
+	const Link = ({
```
sb-dev, line 283: It might be worth moving the Link component to it's own file.

sb-dev: You can to use `handleSelectItem` to update the config state?

sb-dev (approved):

PR #112, opened 2025-01-27
Add log metric filter for source feeds on ingestion

bryophyta:
## What does this change?

- Add a logs metric filter to the log group for the ingestion lambda, which collects the `sourceFeed` from successfully ingested items.
- Also adds the corresponding logging to the lambda handler code.

## How to test

- [x] Deploy to CODE, verify that appropriate metrics are being produced.

sourceFeed metrics, graphed in Grafana

## Accessibility

### DIFF
--- a/ingestion-lambda/src/handler.ts
+++ b/ingestion-lambda/src/handler.ts
@@ -134,6 +134,12 @@ export const main = async (event: SQSEvent): Promise<SQSBatchResponse> => {
 							console.warn(
 								`A record with the provided external_id (messageId: ${messageId}) already exists. No new data was inserted to prevent duplication.`,
 							);
+						} else {
+							console.log({
+								messageId,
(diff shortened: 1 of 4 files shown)


ingestion-lambda/src/handler.ts
```diff
@@ -134,6 +135,12 @@ export const main = async (event: SQSEvent): Promise<SQSBatchResponse> => {
 							console.warn(
 								`A record with the provided external_id (messageId: ${messageId}) already exists. No new data was inserted to prevent duplication.`,
 							);
+						} else {
+							console.log({
```
bryophyta, line 139: I was very tempted to make an abstraction along the lines of

```
function logMetric({ eventType: EventType, payload: CorrespondingPayload }) {
...
}
```

But currently this is the only application so I thought it best to avoid premature abstraction for now..

sb-dev (approved):

PR #127, opened 2025-01-31
Add supplier column to DB

sb-dev:
## What does this change?

Add Supplier column to DB to improve search by supplier.

## Accessibility


bryophyta (approved):
works as expected locally for me 👍 
image

just one small comment below I think

ingestion-lambda/src/handler.ts
```diff
@@ -124,10 +125,12 @@ export const main = async (event: SQSEvent): Promise<SQSBatchResponse> => {
 
 						const snsMessageContent = safeBodyParse(body);
 
+						const supplier = lookupSupplier(snsMessageContent['source-feed']!)
```
bryophyta, line 128: using the `!` assertion here feels a bit odd to me; I think I'd prefer to either explicitly check that `snsMessageContent['source-feed']` is defined before invoking `lookupSupplier`, or else have `lookupSupplier` explicitly handle `undefined` as an input..

not blocking but I'd be interested to hear what you think

sb-dev: Agreed, I might have left it while testing it locally.

PR #133, opened 2025-02-04
More logging on ingestion-lambda and AP poller

bryophyta:
- Log the externalId of successfully ingested items, along with some other metadata.
- Make the namespacing more consistent for eventTypes.
- Use `logger` helper for logging in `apPoller`.

Logging `externalId` and `sqsMessageId` in more places makes it easier to track a given story across lambda, and easier to track the steps within a given lambda, respectively.

logs in ELK showing a single 'externalId' across both the 'reuters_poller_lambda' app and the 'ingestion_lambda' app

## How to test

- [x] Deploy to CODE, verify that the expected logs are showing up.

## Accessibility

### DIFF
--- a/cdk/lib/newswires.ts
+++ b/cdk/lib/newswires.ts
@@ -1,4 +1,4 @@
-import type {Alarms} from '@guardian/cdk';
+import type { Alarms } from '@guardian/cdk';
 import { GuPlayApp, GuScheduledLambda } from '@guardian/cdk';
 import { AccessScope } from '@guardian/cdk/lib/constants';
 import type { NoMonitoring } from '@guardian/cdk/lib/constructs/cloudwatch';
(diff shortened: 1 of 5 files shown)


cdk/lib/newswires.ts
```diff
@@ -153,31 +155,35 @@ export class Newswires extends GuStack {
 			dimensions: { sourceFeed: '$.message.sourceFeed' },
 		});
 
-		const scheduledCleanupLambda = new GuScheduledLambda(this, `ScheduledCleanupLambda-${this.stage}`, {
```
bryophyta, line 156: this block is just auto formatting I believe; will look into why this isn't flagged by CI..

bryophyta: ah, we're not running prettier as part of eslint here. I'll take a look at fixing that today, but in the meantime I think that this change is in line with prettier config so I'm inclined to leave it here for now unless there are objections 👍

ingestion-lambda/src/handler.ts
```diff
@@ -94,16 +100,21 @@ export const main = async (event: SQSEvent): Promise<SQSBatchResponse> => {
...
 					body,
 				}): Promise<OperationResult> => {
+					logger.log({
+						message: `Processing message for ${sqsMessageId}`,
+						eventType: INGESTION_PROCESSING_SQS_MESSAGE_EVENT_TYPE,
+						sqsMessageId,
+					});
 					try {
-						const messageId = messageAttributes['Message-Id']?.stringValue;
+						const externalId = messageAttributes['Message-Id']?.stringValue;
```
bryophyta, line 109: `externalId` helps to differentiate it from the `sqsMessageId`, and is also what we call this value in the database.

sb-dev (approved):

PR #139, opened 2025-02-05
Send to composer

andrew-nowak:
## What does this change?

Adds in a "send to composer" button, like in the old wires. Sends to Composer with a clientside POST, instead of serverside like old wires did. Right now it just sends the [body and a couple of furniture fields](https://github.com/guardian/newswires/pull/139/files#diff-3ad248132691c3faa651f4708b5ecfd8ece58285ffaa838ed42c02dc504429f0R82-R97), but we can improve and add to these as we go. Once sent, the Composer ID is linked into the wire, so we can show a link to the page. In the future we could think about adding an identifier to backlink from Composer to the wire story that created it (and/or a pinboard integration!)

image
image

## How to test

Run on CODE, send to Composer. Does the Composer page open, and a link get generated to get back to it from Newswires?


bryophyta (approved):
This looks good to me, and works as expected in CODE.

The one slight issue I saw is that Composer sometimes says `File out of date please refresh` when you visit it via the "send to composer" link

file out of date screenshot

I'm guessing it's just a slight bug in Composer, but thought it worth flagging.

PR #142, opened 2025-02-07
Add some slack to stalled poller alarms

bryophyta:
## What does this change?

* Refactor the period and threshold calculations for the stalled poller alarms, to make their dependencies on each other clearer.
* Then, update the threshold for fixed pollers to allow some slack. If the poller actually is stalled we should still hear about it within a couple of minutes, but the slack allows room for slight delays in scheduling.

## How to test

- [x] Check the snapshot diffs
- [x] Deploy to CODE; does everything look right?
  - [x] Ran with this config for a few days and stopped getting flaky alarms
  - [x] Reduced concurrency to 0, to stop the lambda from running, and still received an alert within 5 mins.

## Accessibility


cdk/lib/constructs/pollerLambda.ts
```diff
@@ -115,27 +115,34 @@ export class PollerLambda {
...
+				// long polling
+				return {
+					stalledAlarmThreshold: 1,
+					stalledAlarmPeriod: Duration.minutes(3), // long polling (where we expect http requests would timeout in much less than 3 mins)
+				};
+			}
+			// fixed frequency polling
+			const period = Duration.minutes(15); // fixed frequency polling (because 15mins is max delay for SQS)
+			const periodInSeconds = period.toSeconds();
+			const threshold = Math.floor((periodInSeconds / pollerConfig.idealFrequencyInSeconds) * 0.8); // * 0.8 to allow for some slack
```
twrichards, line 129: v. minor, but variable only used once and method call is as descriptive as the variable
```suggestion
			const threshold = Math.floor((period.toSeconds() / pollerConfig.idealFrequencyInSeconds) * 0.8); // * 0.8 to allow for some slack
```

twrichards (approved):
makes sense, good comments - left a v. minor suggestions for brevity

PR #155, opened 2025-02-14
Newswires UI: decode stories text body to remove escaped characters

sb-dev:
## What does this change?

Decode stories text body to remove escaped characters.

## Accessibility

### DIFF
--- a/ingestion-lambda/src/handler.ts
+++ b/ingestion-lambda/src/handler.ts
@@ -72,14 +72,47 @@ const processCategoryCodes = (
 	}
 };
 
+const decodeBodyTextContent = (text: string | undefined): string | undefined =>
+	text
+		?.replace(
+			/\\u([0-9a-fA-F]{4})/g,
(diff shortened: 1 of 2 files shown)


newswires/client/src/WireDetail.tsx
```diff
@@ -21,6 +21,57 @@ import { ComposerConnection } from './ComposerConnection.tsx';
 import { useSearch } from './context/SearchContext.tsx';
 import type { WireData } from './sharedTypes';
 
+interface ArticleProps {
+	bodyTextContent: string;
+}
+
+const decodeBodyTextContent = (text: string): string =>
```
andrew-nowak, line 28: this will fix up the body for the main display, but still leave these characters in any other place they're used (eg. in the brief summary in the search results, or in the content sent to Composer etc.). I think this should be done as part of parsing the API responses (for both the specific item, and the search results!), or might be even better done as part of ingestion - we've already got several bits of code cleaning the input data, it probably makes sense to locate them all together https://github.com/guardian/newswires/blob/main/ingestion-lambda/src/handler.ts

bryophyta: > or might be even better done as part of ingestion

cleaning on ingestion sounds like a good idea to me 👍

PR #159, opened 2025-02-17
Initial fingerpost-based AP world bucket

bryophyta:
## What does this change?

* Add a set of criteria for the AP World bucket that is loosely based on the Fingerpost criteria. General sense-checking seems to suggest that it's catching the right content, and I'd suggest that we should iterate on it as we go if refinements are required (and this should be easier once the initial pass is merged, so everyone can see it).
* To support the new bucket, this branch also adds the ability to query on the `category_codes` column that was added in #150 

## Accessibility

### DIFF
--- a/newswires/app/db/FingerpostWireEntry.scala
+++ b/newswires/app/db/FingerpostWireEntry.scala
@@ -262,7 +262,7 @@ object FingerpostWireEntry
       case Nil => None
       case categoryCodes =>
         Some(
-          sqls"${syn.categoryCodes} && array[${categoryCodes.map(code => sqls"$code")}]::text[]"
+          sqls"${syn.categoryCodes} && ${textArray(categoryCodes)}"
         )
     }
 
@@ -271,7 +271,7 @@ object FingerpostWireEntry
       case categoryCodesExcl =>
         val cce = this.syntax("categoryCodesExcl")
         val doesContainCategoryCodes =
-          sqls"${cce.categoryCodes} && array[${categoryCodesExcl.map(code => sqls"$code")}]::text[]"
+          sqls"${cce.categoryCodes} && ${textArray(categoryCodesExcl)}"
 
         Some(
           sqls"""|NOT EXISTS (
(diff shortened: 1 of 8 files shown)


newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -258,20 +258,49 @@ object FingerpostWireEntry
...
+      case categoryCodes =>
+        Some(
+          sqls"${syn.categoryCodes} && array[${categoryCodes.map(code => sqls"$code")}]::text[]"
+        )
+    }
+
+    val categoryCodesExclQuery = search.categoryCodesExcl match {
+      case Nil => None
+      case categoryCodesExcl =>
+        val cce = this.syntax("categoryCodesExcl")
```
twrichards, line 272: i think it would be more readable if this was inline myself, rather than single use variable with short name

bryophyta: I see your point, although `cce` is actually reused later in the block. I don't understand the query lib very well so am just following the existing pattern in this file tbh. A refactor could be worthwhile (though we should probably check with Andrew, who knows the library better than me) but if so I feel like it'd be better done alongside refactoring the other queries?

newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -258,20 +258,49 @@ object FingerpostWireEntry
...
 
-    // grr annoying but broadly I think subjects and keywords are the same "axis" to search on
-    val keywordsOrSubjectsQuery = (keywordsQuery, subjectsQuery) match {
-      case (Some(kwq), Some(subq)) => Some(sqls"$kwq OR $subq")
-      case _                       => keywordsQuery orElse subjectsQuery
+    val categoryCodesInclQuery = search.categoryCodesIncl match {
+      case Nil => None
+      case categoryCodes =>
+        Some(
+          sqls"${syn.categoryCodes} && array[${categoryCodes.map(code => sqls"$code")}]::text[]"
```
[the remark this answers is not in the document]
bryophyta: ah, I'd missed this before -- thanks! Done now 👍

andrew-nowak (approved):

PR #165, opened 2025-02-20
Fix: More sensible handling of headline and subhead in wires list

bryophyta:
## What does this change?

Feed list: 

* main heading: use `headline` in preference to `slug` for the main heading; only use slug if there is no headline.
* sub heading: instead of using the `headline`, use the `subhead` if it's present, or the first 100 characters of the body text if not. Only show if it's not the same as the heading.

Detail view:

Show the `slug` as part of the headline, but visually de-emphasized unless there is no `headline`.

This seems like the 'right' way to use `slug`, `headline` and `subhead`, and it's more in line with how the agencies render their own content. There are cases where this doesn't work perfectly for our current data (noted below) but it feels like we should try to fix those upstream unless that's not possible.

## Known issues

These are not ideal, but it feels most sensible to try and fix these upstream so that we can use the headline as the main heading, and the subhead (where present) as the subheading.

### AFP

Currently all the AFP stories we're getting from Fip have their subhead set to `JZ`, which is surfaced by this change:
image

### AP

AP stories from Fip seem to have their `headline` and `subhead` switched around.

### PA

When a `headline` is set, it's often repeated at the beginning of the body text, which means that the subheading is repetitive.

## How to test

* Running locally against the CODE db should provide a good sense of how this works with the existing feeds.

## Images

|Before|After|
|-|-|
|image|image|
|image|image|
|image|image|

## Accessibility

### DIFF
--- a/newswires/client/src/Item.tsx
+++ b/newswires/client/src/Item.tsx
@@ -9,6 +9,7 @@ import {
 	EuiSpacer,
 	EuiSplitPanel,
 	EuiSwitch,
+	EuiText,
 	EuiTitle,
 	useGeneratedHtmlId,
 } from '@elastic/eui';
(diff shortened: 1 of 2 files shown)


andrew-nowak (approved):
lgtm! I'm being very nitpicky with the suggestions, feel free to ignore lol

PR #169, opened 2025-02-21
Feed card layout tweaks

bryophyta:
## What does this change?

* Fix a bug in the wire feed card layout, whereby the time stamp was being pushed down to the bottom of the card if the date was also showing (i.e. when it wasn't today's date)
* Move the supplier badge to the right hand side, under the date, to de-emphasise
* Fix a regression whereby the main heading wasn't being rendered in bolder text anymore
* When we have a search highlight to show, don't render other secondary content.

## Images

|Before|After|
|-|-|
|image|image|
|image|image|

## Accessibility

### DIFF
--- a/newswires/client/src/WireItemList.tsx
+++ b/newswires/client/src/WireItemList.tsx
@@ -1,10 +1,8 @@
 import {
 	EuiBadge,
 	EuiButton,
-	EuiHeader,
 	EuiSpacer,
 	EuiText,
-	EuiTitle,
 	useEuiBackgroundColor,
 	useEuiTheme,
 } from '@elastic/eui';
--- a/newswires/client/src/WireItemList.tsx
+++ b/newswires/client/src/WireItemList.tsx
@@ -97,33 +97,35 @@ function decideMainHeadingContent({
 	return 'No headline';
 }
 
-function SecondaryCardContent({
+function MaybeSecondaryCardContent({
 	headline,
 	subhead,
 	bodyText,
 	highlight,
-}: Pick<WireData['content'], 'headline' | 'subhead' | 'bodyText'> & {
+}: WireData['content'] & {
 	highlight: string | undefined;
 }): string | ReactNode | undefined {
 	const theme = useEuiTheme();
 
 	if (highlight && highlight.trim().length > 0) {
 		return (
-			<EuiText
-				size="xs"
-				css={css`
-					margin-top: 0.1rem;
-					padding: 0.1rem 0.5rem;
-					background-color: ${theme.euiTheme.colors.highlight};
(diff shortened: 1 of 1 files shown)


newswires/client/src/WireItemList.tsx
```diff
@@ -150,14 +174,17 @@ const WirePreviewCard = ({
 
 	const mainHeadingContent = decideMainHeadingContent(content);
 
-	const maybeSecondaryCardContent = decideSecondaryCardContent(content);
+	const maybeSecondaryCardContent = SecondaryCardContent({
```
andrew-nowak, line 177: since this now uses a hook, it's a fully fledged component so should be called with jsx (it's probably the same either way, but components can behave oddly when called as functions)

newswires/client/src/WireItemList.tsx
```diff
@@ -1,11 +1,15 @@
 import {
 	EuiBadge,
 	EuiButton,
+	EuiHeader,
```
twrichards, line 4: couple of unused imports, which need cleaning up

newswires/client/src/WireItemList.tsx
twrichards: agree with Andrew, but in my view we should consider calling the component MaybeSecondaryCardContent or SecondaryCardContentIfApplicable, move the <p> into its definition and then instantiate it inline in the final return block of `WirePreviewCard` like... ```<MaybeSecondaryCardContent {...content} highlight={highlight} />```

newswires/client/src/WireItemList.tsx
```diff
@@ -95,11 +99,31 @@ function decideMainHeadingContent({
...
 }
 
-function decideSecondaryCardContent({
+function SecondaryCardContent({
 	headline,
 	subhead,
 	bodyText,
-}: WireData['content']): string | undefined {
+	highlight,
+}: Pick<WireData['content'], 'headline' | 'subhead' | 'bodyText'> & {
```
twrichards, line 107: is there a reason to do the `Pick` here, is there harm in just using the type for the whole object... since that's what is being passed in at the call site.

twrichards (approved):
love before/after of all these changes - nice one! 🙌 
couple of non-blocking comments

newswires/client/src/WireItemList.tsx
bryophyta: Being more explicit about what it was actually using, but I don't feel strongly on this so happy to change 👍

newswires/client/src/WireItemList.tsx
bryophyta: Thanks both! I've done as suggested 👍

PR #174, opened 2025-02-24
Pf/visited link indicator

bryophyta:
## Accessibility


newswires/client/src/urlState.ts
```diff
@@ -51,7 +51,7 @@ export function urlToConfig(location: {
 	} else if (page.startsWith('item/') && page.split('/').length === 2) {
 		return { view: 'item', itemId: page.split('/')[1], query };
 	} else {
-		console.warn(`Page not found: "${page}", so using defaultConfig`);
+		console.log(`Page not found: "${page}", so using defaultConfig`);
```
bryophyta, line 54: Having this as a `warn` was making the logs quite noisy when running tests. Given that we're not storing client-side logs in production, there doesn't seem much benefit to keeping this as a `warn`, as opposed to a `log`.

newswires/client/src/context/SearchContext.tsx
```diff
@@ -110,9 +112,12 @@ export const SearchContext: Context<SearchContextShape | null> =
 	createContext<SearchContextShape | null>(null);
 
 export function SearchContextProvider({ children }: PropsWithChildren) {
-	const [currentConfig, setConfig] = useState<Config>(
+	const [currentConfig, setConfig] = useState<Config>(() =>
 		urlToConfig(window.location),
 	);
+	const [viewedItemIds, setViewedItemIds] = useState<string[]>(() =>
+		loadFromLocalStorage<string[]>('viewedItemIds', z.array(z.string()), []),
+	);
```
twrichards, line 120: as per your excellent segment in Engineering Meet-up, this will be run on every render, and I have a vague memory you'll get warnings if the value passed to initial state varies from render-to-render and even if not, it's probably wasteful, might be worth pulling out the call to `loadFromLocalStorage` into a `useMemo(    , [])` above the set state.

twrichards (approved):
love this feature and the code LGTM (one comment)

I'm a bit unsure/cautious about making the supplier badge grey, I wonder if rather than the multiple visual effects you have denoting _visited_ here, would be simpler as a simple `opacity: 0.5`, as I've tried with dev tools here...
image

... approving as I don't feel too strongly, but I think the opacity approach is slightly more consistent.

bryophyta:
> love this feature and the code LGTM (one comment)
> 
> I'm a bit unsure/cautious about making the supplier badge grey, I wonder if rather than the multiple visual effects you have denoting _visited_ here, would be simpler as a simple `opacity: 0.5`, as I've tried with dev tools here... image
> 
> ... approving as I don't feel too strongly, but I think the opacity approach is slightly easier for me to visually grep.

Thanks! Personally I quite like the look of the 0.5 opacity option that you suggested, but we'd be making it harder to read for some users, which doesn't feel like the right way to go overall..

I was a bit unsure about greying out the supplier badge too, tbh. What do you think of replacing it with a lighter shade when the story has been viewed? e.g.

image

newswires/client/src/context/SearchContext.tsx
bryophyta: Good point, but I think we're okay on this one, because the call to `loadFromLocalStorage` is wrapped in an anonymous function (albeit slightly obscured by the formatting here!) so will only get called once on initial render 👍

sb-dev (approved):

twrichards:
> I was a bit unsure about greying out the supplier badge too, tbh. What do you think of replacing it with a lighter shade when the story has been viewed?

@bryophyta fair enough w.r.t. to accessibility, perhaps we do 0.5 opacity on the badge rather than grey. 

Another thought I had would be to a little unread icon on the ones which haven't been read, a bit like on this mail example https://ui.shadcn.com/examples/mail

bryophyta:
> Another thought I had would be to a little unread icon on the ones which haven't been read, a bit like on this mail example https://ui.shadcn.com/examples/mail

I was thinking of something similar to this, too -- inspired by the github notifications UI which is similar. The reason why I decided against it was because I don't think we want to create the feeling that you _need_ to read everything, and the 'unread' dot might create that vibe. But definitely not closed to the idea. I'm tempted to merge as-is (with the change to desaturate the supplier badge, rather than greyscale it) and then maybe tweak/revisit notification icons later on?

PR #175, opened 2025-02-24
Newswires UI: filter search by date

sb-dev:
## What does this change?

Filter stories by date:
1. Provides a way to filter stories by date using relative or/and absolute dates.
2. By default, it only shows stories ingested during the past two weeks.
3. The URL get updated with only required information (e.g. when the end date is set to `now`, it is not necessary to include in the URL).

### Date Time Component

1. Click on dropdown to get access to the `Quick select` and `Commonly used` menus.
2. Click on date button (showing `Today` in the screenshot) to get access to the advanced date time menu.

#### Select a relative date range
select-relative-date

#### Select an advanced date range
select-advanced-date

### Summary

Summarises the selected date range and allows to reset it.

#### Relative date range

Today:
summary-today

Last 24 hours:
summary-last-24-hours

#### Absolute date range

summary-absolute-date-range

### URL

The date range is reflected in the URL.

#### Relative date range

url-relative-date-range

#### Partial date range

When the end date is set to `now`, it can be omitted from the URL.

url-partial-relative-date-range

#### Absolute date range

url-absolute-date-range

### History

Updates the history and can be combined with other filters.

#### Simple history
history

#### Combined history
history-combined

#### Default date range

screenshot-last-2-weeks

#### Today

screenshot-today

#### Restricted

screenshot-restricted

#### No results

screenshot-no-results

## Accessibility

### DIFF
--- a/newswires/client/src/App.tsx
+++ b/newswires/client/src/App.tsx
@@ -15,6 +15,7 @@ import {
 } from '@elastic/eui';
 import { css } from '@emotion/react';
 import { useSearch } from './context/SearchContext.tsx';
+import { isRestricted } from './dateMathHelpers.ts';
 import { Feed } from './Feed';
 import { Item } from './Item';
 import { SideNav } from './SideNav';
@@ -31,7 +32,7 @@ export function App() {
 		handlePreviousItem,
 	} = useSearch();
 
-	const { view, itemId: selectedItemId } = config;
+	const { view, itemId: selectedItemId, query } = config;
 	const { status } = state;
 
 	const isPoppedOut = !!window.opener;
@@ -74,9 +75,26 @@ export function App() {
 						</p>
 					</EuiToast>
 				)}
+				{isRestricted(query.dateRange?.end) && (
+					<EuiToast
+						title="Restricted"
+						iconType="warning"
+						css={css`
+							border-radius: 0;
+							background: #fdf6d8;
+							position: fixed;
+						`}
+					>
+						<p>
+							Your current filter settings exclude recent updates. Adjust the
+							filter to see the latest data.
+						</p>
+					</EuiToast>
+				)}
 				<div
 					css={css`
-						${status === 'offline' && 'padding-top: 84px;'}
+						${(status === 'offline' || isRestricted(query.dateRange?.end)) &&
+						'padding-top: 84px;'}
 						height: 100%;
 						${(status === 'loading' || status === 'error') &&
 						'display: flex; align-items: center;'}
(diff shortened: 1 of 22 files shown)


newswires/client/src/icons.ts
```diff
@@ -28,10 +31,12 @@ import { icon as popout } from '@elastic/eui/es/components/icon/assets/popout';
 import { icon as refresh } from '@elastic/eui/es/components/icon/assets/refresh';
 import { icon as returnKey } from '@elastic/eui/es/components/icon/assets/return_key';
 import { icon as search } from '@elastic/eui/es/components/icon/assets/search';
+import { icon as sortRight } from '@elastic/eui/es/components/icon/assets/sortRight';
```
bryophyta, line 34: I think we need to add `sortLeft`, too. When I select the 'Absolute' time option I get the following:

`Uncaught Error: Module not found in bundle: ./assets/sortLeft`

newswires/client/src/App.tsx
```diff
@@ -74,9 +75,26 @@ export function App() {
 						</p>
 					</EuiToast>
 				)}
+				{isRestricted(query.dateRange?.end) && (
```
bryophyta, line 78: if both toasts are rendered, one will sit on top of the other: 
image

(I adjusted the opacity of the top one in dev tools for that screenshot)

bryophyta: one option here might be to automatically disable 'auto-refresh' if `isRestricted`, and display the note about not updating next to the auto refresh toggle?

probably only worth doing that if it's fairly quick and easy to implement though, as I suspect we might end up tweaking refresh behaviour before too long...

sb-dev: I tried to automatically disable 'auto-refresh' but it got to messy. I will prioritise the offline message.

newswires/client/src/urlState.ts
```diff
@@ -74,8 +97,37 @@ export const configToUrl = (config: Config): string => {
 	}
 };
 
+const processDateMathRange = (config: Query, useDateTimeValue: boolean) => {
+	if (useDateTimeValue) {
+		if (config.dateRange) {
```
bryophyta, line 102: could we add a comment along the lines of "The API takes fully ISO timestamp strings, so if we're preparing the params for an API call, we need to convert any relative time info to ISO timestamps"

(assuming I'm understanding the logic here correctly!)

bryophyta (approved):
this looks great to me! And really well documented in the PR 🤩 

I've added a couple of comments and suggestions but I don't think any of it is blocking 👍

newswires/client/src/App.tsx
bryophyta: > I tried to automatically disable 'auto-refresh' but it got to messy

got it 👍 I was looking at refactoring the auto refresh state in a separate branch anyway, so if I get round to that I'll see if it'd make it any more straightforward to implement that later

PR #179, opened 2025-02-27
re-enable 'automatic minor version upgrades' for the postgres DB

twrichards:
We were meaning to do this anyway (see the TODO that are now gone) but it also address an FSBP finding. 

NOTE: Unfortunately we still need to leave the version explicitly set to 16.4 (since CFNs default is 16.3 and it can't downgrade), which kind of makes the autoMinorUpgrades problematic since they will force the CDK out of sync with reality 🙄 .


andrew-nowak (approved):
I do strongly dislike how we will fall out of sync, but this seems to be the expected behaviour by cfn & cdk 🤷 

Maybe this will prompt us to attempt an early upgrade to 17, and there very carefully and intentionally specify only the major version number, and let RDS manage the upgrades from there on 🤞

PR #180, opened 2025-02-27
Process AFP category codes on ingestion

bryophyta:
## What does this change?

- Handle incoming AFP subject codes in the same way as AP codes (introduced in #150)
- Fix a small bug in the AP code parsing which would pass through 'empty' tags with no actual code attached (i.e. `iptccat:`) 

## Accessibility

### DIFF
--- a/ingestion-lambda/src/categoryCodes.ts
+++ b/ingestion-lambda/src/categoryCodes.ts
@@ -14,19 +14,21 @@ function partition<T>(
 	return [first, second];
 }
 
-/**
- * We receive AP codes from Fingerpost in the format `prefix:code1+code2+code3:code4+code5`.
- * At the time of writing these are AP category codes, but mislabelled as `iptccat` codes.
- * This function transforms the prefix, and splits the codes into individual category codes.
- */
-function flattenCategoryCodes(categoryCodes: string): string[] {
-	const [prefix, ...codes] = categoryCodes.split(':');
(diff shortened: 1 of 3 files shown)


ingestion-lambda/src/categoryCodes.ts
```diff
@@ -40,19 +43,49 @@ export function processFingerpostAPCategoryCodes(original: string[]): string[] {
...
+
+// example input: "iptccat:HUM+SCI"
+export function processFingerpostAFPCategoryCodes(
+	original: string[],
+): string[] {
+	function flattenCategoryCodes(categoryCodes: string): string[] {
+		const [prefix, ...codes] = categoryCodes.split(':');
+		return codes
+			.flatMap((_) => _.split('+'))
+			.filter((_) => _.trim().length > 0)
```
sb-dev, line 73: The part is duplicated; it might be good to move it to a function:
```
codes
    .flatMap((_) => _.split('+'))
    .filter((_) => _.trim().length > 0)
```

sb-dev (approved):

ingestion-lambda/src/categoryCodes.ts
bryophyta: I did consider that, but then I was thinking that I'd follow the 'rule of three' and abstract it if we see a third application of the same pattern?

PR #187, opened 2025-03-04
Add support for local timezones

sb-dev:
## What does this change?

Ensure that dates are converted to the local timezone when received on the frontend, and to UTC when sent to the backend.

If the UTC offset is not taken into account when filtering stories:

- __Start of BST__: at 1:00 AM GMT (UTC), clocks move forward one hour to 2:00 AM BST (UTC+1), any timestamps between 1:00 and 2:00 AM UTC (which don't map to a local time) will be omitted.

- __End of BST__:  at 2:00 AM BST, clocks move back one hour to 1:00 AM GMT, there will be overlapping stories for the repeated hour.

### Start of BST

daylight-saving-test

### End of BST

daylight-saving-test

We need to extend the time window to retrieve any stories that might have overlapped if the UTC offset wasn’t accounted for.

daylight-saving-test

## Accessibility


newswires/client/src/dateHelpers.ts
```diff
@@ -1,11 +1,22 @@
...
-import moment from 'moment';
+import moment from 'moment-timezone';
 
 export interface TimeRange {
 	start: string;
 	end: string;
 }
 
+export const convertToLocalDate = (timestamp: string): string => {
+	const cleanedTimestamp = timestamp.replace(/\[.*]$/, '');
```
bryophyta, line 10: If I understand the regex correctly then we're using it to discard the timezone information from the timestamp. But this function will only produce the correct local time if the `timestamp` is UTC, won't it? In which case, would it be better to replace `\[Etc\/UTC]$` instead of `\[.*]$`, because we probably want the function to error if we pass a non-UTC timestamp to it.

bryophyta (approved):
Timezones always make my head spin 😵 But the code looks good and the testing I've done locally matches my expectations too!

I've added one question/comment but other than that lgtm!

newswires/client/src/context/SearchReducer.ts
```diff
@@ -1,27 +1,49 @@
...
+	reducer: (state: State, action: Action) => State,
+) => {
+	return (state: State, action: Action): State => {
+		try {
+			return reducer(state, action);
+		} catch (error: unknown) {
+			return {
+				...state,
+				status: 'error',
+				error: error instanceof Error ? error.message : 'Unknown error',
```
bryophyta, line 19:
```suggestion
				error: `Error processing action (${action.type}) ${error instanceof Error ? error.message : 'Unknown error'}`,
```

good to add a bit more context?

PR #200, opened 2025-03-11
Process subjects -> categoryCodes for all incoming wires

bryophyta:
## Accessibility


ingestion-lambda/src/categoryCodes.test.ts
```diff
@@ -32,12 +33,6 @@ describe('processFingerpostAPCategoryCodes', () => {
 		]);
 	});
 
-	it('should pass other codes through untransformed', () => {
```
bryophyta, line 35: I haven't seen any cases where codes include a `+` but we don't want to unpack it into multiple codes, so I think it's reasonable to drop this functionality & test case

sb-dev (approved):

PR #210, opened 2025-03-13
Newswires UI: add story details action buttons

sb-dev:
## What does this change?

A story details action buttons:

- Update the close button to display a cross instead of an arrow
- Move close button to the right
- Add two buttons to navigate stories as an alternative to using keyboard shortcuts

story details action buttons

## Accessibility

### DIFF
--- a/newswires/client/src/Item.tsx
+++ b/newswires/client/src/Item.tsx
@@ -15,10 +15,14 @@ export const Item = ({
 	error,
 	itemData,
 	handleDeselectItem,
+	handlePreviousItem,
+	handleNextItem,
 }: {
 	error: string | undefined;
 	itemData: WireData | undefined;
 	handleDeselectItem: () => void;
+	handlePreviousItem: () => void;
+	handleNextItem: () => void;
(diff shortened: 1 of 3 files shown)


newswires/client/src/Item.tsx
```diff
@@ -35,13 +39,21 @@ export const Item = ({
...
-								justifyContent="flexEnd"
-								alignItems="center"
-							></EuiFlexGroup>
+							<EuiButtonIcon
+								iconType="arrowRight"
+								onClick={handleNextItem}
+								aria-label="Next story"
+							/>
+							<EuiFlexGroup justifyContent="flexEnd" alignItems="center">
+								<EuiButtonIcon
```
bryophyta, line 51: I wonder if we want to automatically move focus into this panel when it's opened, possibly onto the close button?

I think we had some similar behaviour before but I seem to have broken it at some point... 🤔

bryophyta (approved):
lgtm and works as expected when tested locally! 🥳

PR #215, opened 2025-03-17
ingestion: fix bad control characters

andrew-nowak:
## What does this change?

We receive a bunch of input documents where the body text field contains a literal tab or newline character (rather than the usual "\t" or "\n" escaped versions), which are invalid in JSON. Make a best effort to fix them up, but this is a little tricky as these whitespace characters are valid outside of JSON strings (ie. as indentation around the object literal).

Fortunately they can be safely replaced with " "s, in both locations!

## How to test

Tested on CODE, where the remaining 234 messages on the DLQ were redriven through the lambda running this updated code with success.

More messages make it onto newswires without parsing failures

### DIFF
--- a/ingestion-lambda/src/handler.ts
+++ b/ingestion-lambda/src/handler.ts
@@ -121,10 +121,11 @@ export const safeJsonParse = (body: string): Record<string, unknown> => {
 		if (e instanceof SyntaxError && isBadControlChar(e)) {
 			console.warn('Attempting to strip unescaped tab chars');
 			// technically, the bad control char warning could appear for any control char that remains unescaped
-			// inside a JSON string; but in practice we've only seen this for tab characters so far. If other control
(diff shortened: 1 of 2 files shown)


ingestion-lambda/src/handler.ts
```diff
@@ -95,27 +99,13 @@ export const decodeBodyTextContent = (
 		})
 		.replace(/\n/g, '<br />');
 
-export const safeBodyParse = (body: string): IngestorInputBody => {
+export const safeJsonParse = (body: string): Record<string, unknown> => {
```
andrew-nowak, line 102: this needed splitting into two steps, otherwise the "fall back" step which cleans the curly quotes and/or fixes control chars will not have the preprocessed keywords or body content

sb-dev (approved):

PR #233, opened 2025-03-28
UI: update beta notice

sb-dev:
## What does this change?

- Update beta notice text
- Add a beta notice modal 
- Update the mailto link to work with gmail.

## Accessibility

### DIFF
--- a/newswires/client/src/App.tsx
+++ b/newswires/client/src/App.tsx
@@ -1,22 +1,32 @@
 import {
 	EuiBetaBadge,
 	EuiButton,
+	EuiButtonEmpty,
 	EuiEmptyPrompt,
 	EuiHeader,
 	EuiHeaderSection,
 	EuiHeaderSectionItem,
+	EuiIcon,
 	EuiLink,
+	EuiModal,
+	EuiModalBody,
+	EuiModalFooter,
+	EuiModalHeader,
+	EuiModalHeaderTitle,
 	EuiPageTemplate,
 	EuiProvider,
 	EuiResizableContainer,
 	EuiShowFor,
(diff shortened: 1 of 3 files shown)


newswires/client/src/App.tsx
```diff
@@ -81,6 +105,36 @@ export function App() {
 					max-height: 100vh;
 				`}
 			>
+				{displayDisclaimer && (
+					<EuiModal
+						aria-labelledby={'Use with caution'}
```
bryophyta, line 110: If we're setting the aria label inline then the property should be `aria-label` instead. I'd be inclined to keep it as `aria-labelledby` though, and generate an id for the EuiModalHeaderTitle that can be passed as the value, so that the aria label will automatically match the value of the header title

newswires/client/src/App.tsx
```diff
@@ -81,6 +105,36 @@ export function App() {
...
+								<FeedbackContent />
+							</EuiText>
+						</EuiModalBody>
+
+						<EuiModalFooter>
+							<EuiButtonEmpty onClick={() => dismissDisclaimer()}>
+								Close
+							</EuiButtonEmpty>
+							<EuiButton onClick={() => dismissDisclaimer(true)} fill>
+								Acknowledge
```
bryophyta, line 133: I'm inclined to make it explicit that this means the pop-up won't be shown again.. would `Acknowledge (don't show again)` be too verbose..?

bryophyta (approved):
Looks good to me -- clearer wording and works as expected!

The aria label for the modal is slightly broken, and I have a couple of small suggestions about wording, but overall lgtm!

PR #235, opened 2025-04-01
Minor agencies world

sb-dev:
## What does this change?

Add minor agencies world: since minor agencies don’t have category codes, infer the UK region category code using known UK locations, and exclude the category code from the search preset.

## Accessibility

### DIFF
--- a/ingestion-lambda/src/categoryCodes.ts
+++ b/ingestion-lambda/src/categoryCodes.ts
@@ -1,3 +1,6 @@
+import nlp from 'compromise';
+import {lexicon, ukPlaces} from "./ukPlaces";
+
 interface CategoryCode {
 	prefix: string;
 	code: string;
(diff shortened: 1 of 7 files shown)


ingestion-lambda/src/categoryCodes.ts
```diff
@@ -110,3 +112,32 @@ export function processUnknownFingerpostCategoryCodes(
...
+		return undefined;
+	}
+
+	const places = (rawPlaces as string[])
+		.flatMap((place) => place.split(/[,\n]/))
+		.map((place) => place.trim().toLowerCase())
+		.filter((place) => !!place && place.length > 0);
+
+	const isUk = places.some((place) =>
+		ukPlaces.some((ukPlace) => place.includes(ukPlace)),
```
bryophyta, line 139: Matching on `includes` means that the following test case fails:

```
it('should return undefined when only non-UK places are mentioned', async () => {
	const content = 'Kingsford Smith, Bradfield, Sydney, and Warringah in New South Wales and Fadden in Queensland are the five most expensive electorates, data shows.';
	expect(await inferRegionCategoryFromText(content)).toEqual(undefined);
});
```

Could we try doing exact match, or are there other edge cases that we'd get wrong if we do that? I know there will always be edge cases for us here, so not blocking 👍

bryophyta (approved):
Small comment on place-matching logic but not blocking 👍 

Looks like a good pragmatic approach to classification to me! And it looks like `compromise` is sufficiently light-weight that it shouldn't slow down ingestion significantly.

It'll be interesting to see how it performs against some actual stories, and should be easy to iterate down the line if we need to make any changes 🥳

ingestion-lambda/src/categoryCodes.ts
sb-dev: It looks like an edge case to keep in mind. I was thinking to try out this iteration and make further updates to it later if needed.

PR #236, opened 2025-04-03
UI: move ticker button

sb-dev:
## What does this change?

- Add ticker buttons next to the supplier filter in the side nav (shows on hover)
- Add a ticker button in the search summary (include a tooltip)
- Add custom tooltip (as the Elastic UI tooltip doesn't work)
- Replace the ticker button in the search summary with a beacon in ticker windows (the beacon becomes red when the app goes offline)

## Images

Supplier filter:
Supplier filter

Hover state:
Hover state

Search summary:
Search summary

Ticker window:
Ticker window

Offline ticker window:
Offline ticker window

## Accessibility

### DIFF
--- a/newswires/client/src/SideNav.tsx
+++ b/newswires/client/src/SideNav.tsx
@@ -1,3 +1,4 @@
+import type { EuiPinnableListGroupItemProps } from '@elastic/eui';
 import {
 	EuiBadge,
 	EuiBadgeGroup,
@@ -8,6 +9,7 @@ import {
 	EuiIcon,
 	EuiListGroup,
 	EuiListGroupItem,
+	EuiPinnableListGroup,
 	EuiSwitch,
 	EuiText,
 	useEuiTheme,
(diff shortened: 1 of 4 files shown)


bryophyta (commented):
I've been looking at [the MDN page for tooltips](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tooltip_role), and it looks like we should do a couple of extra things:

1. `role='tooltip'` on the tooltip itself
2. make the tooltip target (i.e. the button in most cases) be `aria-describedby` the tooltip
3. make the tooltip be dismissable by pressing the `escape` key

Not sure what the best way is to inject the `aria-describedby` attribute onto the target. I found [this addAriaDescribedby function](https://github.com/guardian/flexible-content/blob/1cc89f8c7e7545783fc239c79aeecb55b8aa4532/flexible-frontend/src/components/Tooltip.tsx#L119-L124) in another codebase, but they're using a slightly different approach so it might not fit here.

Probably shouldn't block this PR so if it's a significant piece of work we could add it in a separate branch, but feels like we should aim to conform to the MDN spec here?

newswires/client/src/SideNav.tsx
```diff
@@ -134,6 +159,38 @@ export const SideNav = () => {
...
+						height: 1.5rem;
+
+						margin-right: 12px;
+						background-color: ${isActive ? colour : 'transparent'};
+					`}
+				/>
+			),
+			color: isActive ? 'primary' : 'subdued',
+			pinnable: false,
+			extraAction: {
```
bryophyta, line 181: generally this looks great! but when hovering over a non-selected supplier the contrast between the popout button and the supplier button is very faint -- could we make the contrast easier to see?
image

bryophyta (approved):
Looks great to me, and really great to have working tooltips too! A couple of small comments, and one larger one about the tooltip accessibility. The latter could be split off into a separate PR if it proves to be a big job?

PR #241, opened 2025-04-04
UI: make Alert messages smaller

sb-dev:
## What does this change?

- Reduce the height of alert messages. 

## Images

Restricted search:
Restricted search

Offline:
Offline

## Accessibility

### DIFF
--- a/newswires/client/src/App.tsx
+++ b/newswires/client/src/App.tsx
@@ -66,6 +66,7 @@ const Alert = ({
 				border-radius: 0;
 				background: #fdf6d8;
 				position: fixed;
+				z-index: 1000;
 			`}
 		></EuiToast>
 	</div>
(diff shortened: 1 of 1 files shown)


newswires/client/src/App.tsx
```diff
@@ -39,23 +39,34 @@ import { configToUrl, defaultQuery } from './urlState';
...
 }) => (
-	<EuiToast
-		title={title}
-		iconType={icon}
+	<div
 		css={css`
-			border-radius: 0;
-			background: #fdf6d8;
-			position: fixed;
+			.header-only span {
```
bryophyta, line 48: The screenshots look great, but I feel a bit uneasy about using `!important` for styling here. Is there any reasonable way of avoiding it? If not, could we add an inline comment explaining why we think it's necessary?

sb-dev: I removed the `!important` annotation and narrowed down the selector.

bryophyta (approved):
lgtm!

PR #260, opened 2025-04-15
add newswires favicon

andrew-nowak:
## What does this change?

Replace the default/missing favicon with this beauty: 

image
courtesy of @paperboyo


bryophyta (approved):
fab!

PR #262, opened 2025-04-16
Fix copy & paste of body text

bryophyta:
## What does this change?

Copy & paste was often bringing in visually hidden fields, and list styling, which is not ideal.

On reflection, there's no great need for the byline, body text and abstract to be rendered as part of the description list, because they're the main content of the item. So a simple solution is to pull them out of it.

## Images

|Before|After|
|-|-|
|image|image|

## Accessibility

### DIFF
--- a/newswires/client/src/WireDetail.tsx
+++ b/newswires/client/src/WireDetail.tsx
@@ -10,7 +10,6 @@ import {
 	EuiDescriptionListTitle,
 	EuiFlexGroup,
 	EuiFlexItem,
-	EuiScreenReaderOnly,
 	EuiSpacer,
 	EuiText,
 	EuiTitle,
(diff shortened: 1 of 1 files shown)


newswires/client/src/WireDetail.tsx
```diff
@@ -276,65 +282,47 @@ export const WireDetail = ({
 					</h3>
 					<EuiSpacer size="s" />
 					{ednote && <EuiCallOut size="s" title={ednote} color="success" />}
-					<EuiSpacer size="s" />
+					<EuiSpacer size="m" />
```
andrew-nowak, line 285: bit tangential to this PR, but should one of these spacers be inside the `{ednote` conditional block? otherwise I imagine we end up with a double spacer

bryophyta: good call, cheers! I've fixed that now 👍

PR #263, opened 2025-04-17
eslint 9 upgrade

andrew-nowak:
## What does this change?

Attempts the eslint 9 upgrade


newswires/client/eslint.config.mjs
```diff
@@ -0,0 +1,21 @@
+import guardian from '@guardian/eslint-config';
+import storybook from 'eslint-plugin-storybook';
+import newswiresConfig from '../../eslint.config.mjs';
+
+export default [
+	...newswiresConfig,
+	...guardian.configs.react,
```
bryophyta, line 7: we seem to be losing some rules in the client project in the new config.

e.g. I deleted the import of `EuiProvider` in App.tsx and on `main` eslint gives the following:

`error    'EuiProvider' is not defined                                                  react/jsx-no-undef`

but this doesn't seem to be picked up by eslint in this branch. I haven't been able to track down yet where the rule was coming from in `main`, so possible that this is due to a change in the Guardian configs? But it looks like we were adding the `'plugin:react/jsx-runtime'` plugin explicitly in the previous client eslint config, so I wonder whether we need to find a way to re-add it here?

I'll try and dig into this a bit more but wanted to jot this down before the long weekend as I'll forget otherwise 🙃 And/or maybe this rings a bell already?

andrew-nowak: https://github.com/guardian/csnx/pull/2079

andrew-nowak: Resolved in @guardian/eslint-config v12, 8dedd58

newswires/client/eslint.config.mjs
```diff
@@ -0,0 +1,25 @@
...
+	...storybook.configs['flat/recommended'],
+	{
+		rules: {
+      "@typescript-eslint/switch-exhaustiveness-check": [
+        "error",
+        {considerDefaultExhaustiveForUnions: true},
```
andrew-nowak, line 13: I had to add this in as there's a few bits of code that fall foul of the default for this rule, which does not consider a `default` case as exhausting all the options for the switch. We could consider coming back and refactoring to not use `defaults` and turning this option off...

bryophyta: seems v sensible to me 👍

bryophyta (approved):
This looks great, thanks for this!

The one thing I spotted was that we seem to be losing the prettier config in the client project. I've opened a PR into this branch to add it back in (or, to add `prettier/recommended`, because I can't work out how to get the `@guardian/prettier` config to play with the new eslint config style) https://github.com/guardian/newswires/pull/308

But if it doesn't feel apt as part of this PR then feel free to ignore. 

LGTM!

PR #268, opened 2025-04-17
UI: fix feed item line break

sb-dev:
## What does this change?

Fix feed item line break.

## Images

Before:
Line break issue

After:
Fixed

## Accessibility


bryophyta (approved):
great catch!

paperboyo:
Oh I don’t think this is quite correct to break words like this (in random places and without hyphens). If we want more breaking, we could maybe use CSS hyphenation controls ([auto](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens), there are also new, fancy, [additions](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphenate-limit-chars))? Let’s chat.

PR #270, opened 2025-04-18
Add horizontal/vertical split toggle

bryophyta:
## What does this change?

- Refactor the resizable panels to make it easier to pass a direction prop (horizontal/vertical)
- Add a new setting to allow users to toggle panel direction
- Also save the preferred proportions of the two panels in localstorage, for each orientation, so that it will re-open in the same layout
- On small screens:
    - For standard view, still open items separate from the feed (no split)
    - For ticker view, always open items in a split below the feed, regardless of user preference for direction

image

## Images

|Horizontal|Vertical|
|-|-|
|**Desktop breakpoints**||
|image|image|
|**Tablet breakpoints**||
|image|image|
|**Ticker mode**||
|image|image|

## Accessibility

### DIFF
--- a/newswires/client/src/App.tsx
+++ b/newswires/client/src/App.tsx
@@ -123,12 +123,13 @@ const ResizableContainer = ({
 					>
 						{Feed}
 					</EuiResizablePanel>
-					<EuiResizableButton accountForScrollbars={'before'} />
+					<EuiResizableButton accountForScrollbars={'both'} />
 					<EuiResizablePanel
 						id={secondPanelId}
 						minSize="20%"
 						initialSize={sizes[secondPanelId]}
 						className="eui-yScroll"
+						paddingSize="none"
 					>
 						{Item}
 					</EuiResizablePanel>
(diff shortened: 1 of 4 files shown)


newswires/client/src/App.tsx
```diff
@@ -188,6 +249,7 @@ export function App() {
 						  }
 						`}
 						height: 100%;
+						max-height: 100vh;
```
bryophyta, line 252: this seems to be needed for the vertical split, although I couldn't quite work out why this wasn't handled by the restrictions already put on the parent element..

newswires/client/src/App.tsx
```diff
@@ -262,35 +324,23 @@ export function App() {
...
 							<EuiShowFor sizes={['xs', 's']}>
-								{view === 'item' ? <ItemData id={selectedItemId} /> : <Feed />}
+								{view === 'item' &&
+									(isPoppedOut ? (
+										<ResizableContainer
+											Feed={<Feed />}
+											Item={<ItemData id={selectedItemId} />}
+										/>
+									) : (
+										<ItemData id={selectedItemId} />
```
bryophyta, line 334: for now keeping the pattern of opening items on their own for small screens, but allowing split view if it's a small screen and open in ticker mode

sb-dev (approved):

PR #276, opened 2025-04-23
Fix: header for smaller screens

bryophyta:
## What does this change?

1. Reduce the horizontal space needed for the header elements when rendering in smaller breakpoints:
    - Icon-only button for the new ticker button
    - Icon-only visual rendering of the beta badge
2. Use 'compact' mode for the date picker (not related to the header changes, but felt like a small enough change that it was reasonable to bundle in here)

---

New layout on smaller screens:

image

---

Comparison of non-compact (left) vs. proposed compact (right) versions of the date picker:

image

### DIFF
--- a/newswires/client/src/App.tsx
+++ b/newswires/client/src/App.tsx
@@ -2,6 +2,7 @@ import {
 	EuiBetaBadge,
 	EuiButton,
 	EuiButtonEmpty,
+	EuiButtonIcon,
 	EuiEmptyPrompt,
 	EuiFlexGroup,
 	EuiHeader,
@@ -37,6 +38,7 @@ import { FeedbackContent } from './FeedbackContent.tsx';
 import { ItemData } from './ItemData.tsx';
 import { SettingsMenu } from './SettingsMenu.tsx';
 import { SideNav } from './SideNav';
+import { Tooltip } from './Tooltip.tsx';
 import { configToUrl, defaultQuery } from './urlState';
 
 const Alert = ({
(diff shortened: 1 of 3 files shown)


newswires/client/src/App.tsx
```diff
@@ -217,15 +219,30 @@ export function App() {
...
-												size="s"
-												css={css`
-													margin-left: 8px;
-												`}
-											/>
+											<EuiShowFor sizes={['xs', 's']}>
+												<EuiBetaBadge
+													iconType={'beaker'}
+													label="Currently under construction"
+													aria-label="(Under construction)"
```
bryophyta, line 226: aria-label is required here because for some reason the label & title props don't get put into the accessibility tree otherwise, so will be missing from the heading text for screen readers. 

I also added the same aria-label to the badge for larger screens below (although the label does get included in the accessibility tree there because it's rendered as visible text on the page) so that the heading is standardised across the breakpoints

newswires/client/src/App.tsx
```diff
@@ -238,23 +255,49 @@ export function App() {
...
-												'_blank',
-												'popout=true,width=400,height=800,top=200,location=no,menubar=no,toolbar=no',
-											)
-										}
-									>
-										New ticker
-									</EuiButton>
+									<EuiShowFor sizes={['xs', 's']}>
+										<Tooltip
+											tooltipContent={'Open ticker in a new window'}
```
sb-dev, line 260: The tooltip for the ticker button in the search summary says `Open new ticker`. It would be nice to make both messages consistent.

newswires/client/src/App.tsx
```diff
@@ -238,23 +255,49 @@ export function App() {
...
+												}
+											/>
+										</Tooltip>
+									</EuiShowFor>
+									<EuiShowFor sizes={['m', 'l', 'xl']}>
+										<EuiButton
+											size="s"
+											iconType={'popout'}
+											onClick={() =>
+												window.open(
```
sb-dev, line 287: Can be moved to a common ticker function.

bryophyta: good call, am implementing here: https://github.com/guardian/newswires/pull/279/files

sb-dev (approved):

PR #283, opened 2025-04-25
Scroll to top of item when new item is opened

bryophyta:
## What does this change?

This is something a couple of people have asked for, and it feels like sensible behaviour.

## Accessibility

### DIFF
--- a/newswires/client/src/Item.tsx
+++ b/newswires/client/src/Item.tsx
@@ -8,7 +8,7 @@ import {
 	EuiSwitch,
 	useIsWithinBreakpoints,
 } from '@elastic/eui';
-import { useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { type WireData } from './sharedTypes';
 import { Tooltip } from './Tooltip.tsx';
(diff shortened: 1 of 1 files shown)


newswires/client/src/Item.tsx
```diff
@@ -29,6 +29,23 @@ export const Item = ({
 	const [isShowingJson, setIsShowingJson] = useState<boolean>(false);
 	const isSmallScreen = useIsWithinBreakpoints(['xs', 's']);
 
+	const headingRef = useRef<HTMLHeadingElement>(null);
+
+	// scroll to heading when a new item is selected
+	useEffect(() => {
+		if (headingRef.current) {
+			headingRef.current.scrollIntoView({
+				behavior: 'smooth',
```
sb-dev, line 38: We can probably skip the smooth transition because users will already expect the story to be at the top by default.

bryophyta: sounds good, I've updated the commit removing this prop 👍

sb-dev (approved):

PR #288, opened 2025-04-29
UI: fix item list scrolling

sb-dev:
## What does this change?

When returning to the list from a story don't jump to top.

## Accessibility

### DIFF
--- a/newswires/client/src/App.tsx
+++ b/newswires/client/src/App.tsx
@@ -295,27 +295,33 @@ export function App() {
 					{status !== 'error' && (
 						<>
 							<EuiShowFor sizes={['xs', 's']}>
-								{view === 'item' &&
-									(isPoppedOut ? (
-										<ResizableContainer
-											Feed={<Feed />}
-											Item={<ItemData id={selectedItemId} />}
-											directionOverride={'vertical'}
-										/>
-									) : (
-										<ItemData id={selectedItemId} />
-									))}
-								{view !== 'item' && <Feed />}
-							</EuiShowFor>
-							<EuiShowFor sizes={['m', 'l', 'xl']}>
-								{view === 'item' ? (
+								{isPoppedOut && (
 									<ResizableContainer
 										Feed={<Feed />}
-										Item={<ItemData id={selectedItemId} />}
+										Item={
+											selectedItemId ? (
+												<ItemData id={selectedItemId} />
+											) : undefined
+										}
+										directionOverride={'vertical'}
 									/>
-								) : (
-									<Feed />
 								)}
+
+								{view === 'item' && !isPoppedOut && (
+									<ItemData id={selectedItemId} />
+								)}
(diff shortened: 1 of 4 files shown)


newswires/client/src/WireItemList.tsx
```diff
@@ -26,11 +26,17 @@ export const WireItemList = ({
 	wires: WireData[];
 	totalCount: number;
 }) => {
-	const { config, loadMoreResults } = useSearch();
+	const { config, loadMoreResults, previousItemId } = useSearch();
 
 	const [isLoadingMore, setIsLoadingMore] = useState<boolean>(false);
 
-	const selectedWireId = config.itemId;
+	const [selectedWireId, setSelectedWireId] = useState<string | undefined>(
```
bryophyta, line 33: why does `selectedWireId` need to be a piece of state here? shouldn't the component get re-rendered anyway when we change `config.itemId`?

newswires/client/src/WireItemList.tsx
```diff
@@ -58,6 +64,9 @@ export const WireItemList = ({
 								isFromRefresh={isFromRefresh}
 								highlight={highlight}
 								selected={selectedWireId == id.toString()}
+								selectedId={selectedWireId?.toString()}
```
bryophyta, line 67: this prop doesn't seem to be used in the component

newswires/client/src/App.tsx
```diff
@@ -295,27 +295,33 @@ export function App() {
...
 									/>
-								) : (
-									<Feed />
 								)}
+
+								{view === 'item' && !isPoppedOut && (
+									<ItemData id={selectedItemId} />
+								)}
+
+								{view !== 'item' && <Feed />}
```
bryophyta, line 314: with the current logic, when `isPoppedOut` and `view !== 'item'` are both true the `<Feed>` component will be rendered twice (once here, and once in the resizable container)
```suggestion
								{view !== 'item' && !isPoppedOut && (<Feed />)}
```

newswires/client/src/WireItemList.tsx
```diff
@@ -162,21 +184,32 @@ const WirePreviewCard = ({
...
-				behavior: 'smooth',
-				block: 'nearest',
-			});
+			scrollElementIntoView(ref.current);
 		}
 	}, [selected]);
 
+	useEffect(() => {
+		if (
+			isSmallScreen &&
```
bryophyta, line 204: is this meant to target the case of the regular mobile view only? if so, we need to add `!isPoppedOut` to the list of conditions

my feeling is that we should add a comment to this useEffect hook to explain the intention, as there are quite a few conditions in play so it's not immediately obvious to a reader unless they work through exactly what conditions would trigger the scrolling behaviour

bryophyta (approved):
this looks great to me! 🥳

PR #293, opened 2025-05-12
UI: add jump to top button

sb-dev:
## What does this change?

Add jump to top button:

- Accumulate counts of newly loaded stories and reset when scrolling up
- Show/hide the button based on scroll offset
- Recalculate position on show, scroll, or resize

## Images

Large screen feed only:
Large screen feed only

Large screen with opened story:
Large screen with opened story

Large screen with story opened below feed:
Large screen with story opened below feed

Small screen feed only:
Small screen feed only

Ticker screen feed only:
Ticker screen feed only

Ticker screen feed wit story opened below feed:
Ticker screen feed wit story opened below feed

## Accessibility

### DIFF
--- a/newswires/client/src/App.tsx
+++ b/newswires/client/src/App.tsx
@@ -285,7 +285,7 @@ export function App() {
 							<EuiShowFor sizes={['xs', 's']}>
 								{isPoppedOut && (
 									<ResizableContainer
-										Feed={<Feed />}
+										Feed={Feed}
 										Item={
 											selectedItemId ? (
 												<ItemData id={selectedItemId} />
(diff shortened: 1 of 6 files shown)


bryophyta (approved):
this looks good to me!

I spotted a couple of edge cases, e.g. if you toggle from 'show item below' to 'show item next to feed' the story count seems to reset and the button is briefly rendered in slightly the wrong place until something changes. But I suspect that these are due to the parent container being re-rendered, in which case it might not be worth the time to fix these upfront.

It works well in my testing, and adds useful functionality, so I'd be inclined to merge and see what users think 😃

PR #300, opened 2025-05-16
UI: move search box to page header

sb-dev:
## What does this change?

Move the search box to the page header:
- Show the search box on large screens
- Wrap the search box into a popover on small screens

## Images

Large screen:
search box for large screens

Small screen:
search box for small screens

## Accessibility


bryophyta:
For larger screens, do you think we could align the search field with the start of the main body column (as pictured)? I think it would look good, but not sure if the layout gives us a straightforward way of doing it or not. (Probably not worth it, if not)

image

newswires/client/src/App.tsx
```diff
@@ -238,13 +240,25 @@ export function App() {
 								</EuiHeaderSectionItem>
 							</EuiHeaderSection>
 
+							<EuiHeaderSectionItem>
+								<EuiShowFor sizes={['l', 'xl']}>
+									<SearchBox width={'580px'} />
```
bryophyta, line 245: is there a reason why we couldn't control the width by wrapping `<SearchBox>` in a container? It doesn't feel quite right to me to add a new prop for just the one case


e.g. 

```
<div
	css={css`
		width: 580px;
	`}
>
	<SearchBox/>
</div>
```

bryophyta:
Below about 440px screen width I'm seeing the header bar getting cut off rather than resizing.
 

Extra-small breakpoint screenshot:

image


I think we should look at cutting horizontal space down a bit in the header anyway (e.g. replace the `Newswires` bit with just the logo, perhaps) but it would be good to avoid a regression here in the meantime.. If there isn't room at the moment, should we reinstitute the toggle button just for the `xs` breakpoint for now? (Looks like it's fine at the bottom of the `s` breakpoint to me)


Small breakpoint screenshot:

image

newswires/client/src/SearchMenu.tsx
```diff
@@ -0,0 +1,34 @@
+import { EuiButtonIcon, EuiPopover } from '@elastic/eui';
+import { useState } from 'react';
+import { SearchBox } from './SearchBox.tsx';
+
+export const SearchMenu = () => {
```
bryophyta, line 5: as mentioned in other comment, I think we might still want to use this for `xs` breakpoints, but if we don't then it looks like this component is unused atm?

paperboyo:
> should we reinstitute the toggle button just for the `xs` breakpoint for now

Given some phones may well have logical pixel width triggering this… maybe easier would be to drop beta icon? (just trying to avoid adding a popout?) But yeah, best would be to replace Newswires + beta icon with a logo (favicon).

newswires/client/src/context/KeyboardShortcutsContext.tsx
```diff
@@ -94,10 +96,23 @@ export function useKeyboardShortcuts() {
...
 export const StopShortcutPropagationWrapper = ({
 	children,
 }: {
-	children: React.ReactNode;
+	children: React.ReactElement<HasKeyUpCapture>;
 }) => {
-	return <div onKeyUp={stopShortcutPropagation}>{children}</div>;
+	const existingCapture = children.props.onKeyUpCapture;
+
+	const handleCapture: React.KeyboardEventHandler<HTMLElement> = (e) => {
```
bryophyta, line 110: 👍👍

bryophyta (approved):
I've added a comment about `xs` breakpoints that I think we should address, but once that's addressed this lgtm!

sb-dev:
> > should we reinstitute the toggle button just for the `xs` breakpoint for now
> 
> Given some phones may well have logical pixel width triggering this… maybe easier would be to drop beta icon? (just trying to avoid adding a popout?) But yeah, best would be to replace Newswires + beta icon with a logo (favicon).

We could move the app title and beta badge to the side nav?

Side nav

Main screen

bryophyta:
> > > should we reinstitute the toggle button just for the `xs` breakpoint for now
> > 
> > 
> > Given some phones may well have logical pixel width triggering this… maybe easier would be to drop beta icon? (just trying to avoid adding a popout?) But yeah, best would be to replace Newswires + beta icon with a logo (favicon).
> 
> We could move the app title and beta badge to the side nav?
> 
> Side nav Main screen

that works for me for now!

newswires/client/src/App.tsx
```diff
@@ -187,87 +189,83 @@ export function App() {
 				>
 					{!isPoppedOut && (
 						<EuiHeader position="fixed">
-							<EuiHeaderSection>
+							<EuiHeaderSection side={'left'}>
 								<EuiHeaderSectionItem>
 									<SideNav />
 								</EuiHeaderSectionItem>
+								<EuiShowFor sizes={['s', 'm', 'l', 'xl']}>
```
bryophyta, line 196: this leaves the `xs` page layout without a `h1` landmark heading, unless the menu is opened.

as we've discussed, this seems reasonable overall given the space constraints, and likely expectations of users with narrow screens, but it feels like we should probably make sure there's always a `h1` on the page even if it's visually hidden, so perhaps we could add something like:

```
<ShowFor sizes={['xs'])>
    {!sideNavIsOpen && <ScreenReaderOnly>
        <h1>Newswires (under construction)</h1>
    </>}
</>
```

bryophyta (approved):
just added [one comment](https://github.com/guardian/newswires/pull/300/files#r2108934014) about making sure there's a h1 present on x-small screens, but aside from that this all looks good to me!

PR #302, opened 2025-05-19
Fix search badge toggle behaviour

bryophyta:
## What does this change?

- Fix a bug whereby dismissing a category code search badge wasn't removing the code from the search config.
- As part of addressing this, refactor the SearchSummary component to add a bit more type safety, to hopefully reduce the chance of this happening again, if new badges are created, or old ones are renamed.

image

## Accessibility

### DIFF
--- a/newswires/client/.eslintrc.cjs
+++ b/newswires/client/.eslintrc.cjs
@@ -39,6 +39,7 @@ module.exports = {
 				caughtErrorsIgnorePattern: '^_',
 			},
 		],
+		"@typescript-eslint/switch-exhaustiveness-check": "error",
 		'prettier/prettier': 'warn',
 		'react/no-unknown-property': ['error', { ignore: ['css'] }],
 		'no-restricted-syntax': [
--- a/newswires/client/src/SearchSummary.tsx
+++ b/newswires/client/src/SearchSummary.tsx
@@ -30,12 +30,20 @@ const presetLabel = (preset: string) => {
 	}
 };
 
+type SearchTermBadgeLabel =
+	| 'Search term'
+	| 'Time range'
+	| 'Preset'
+	| 'Supplier'
+	| 'Category'
+	| '(NOT) Category';
+
 const Summary = ({
 	searchSummaryLabel,
 }: {
 	searchSummaryLabel: string | boolean;
 }) => {
-	const { config, handleEnterQuery } = useSearch();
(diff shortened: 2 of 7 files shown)


newswires/client/.eslintrc.cjs
```diff
@@ -39,6 +39,7 @@ module.exports = {
 				caughtErrorsIgnorePattern: '^_',
 			},
 		],
+		"@typescript-eslint/switch-exhaustiveness-check": "error",
```
bryophyta, line 42: I added this because I'd quite like it, and it doesn't introduce any new errors, so it feels like a pretty reasonable thing to enforce. But I'm v open to other perspectives if people think there are situations where we wouldn't want to enforce exhaustiveness of `switch` cases

newswires/client/src/SearchSummary.tsx
```diff
@@ -30,12 +30,20 @@ const presetLabel = (preset: string) => {
 	}
 };
 
+type SearchTermBadgeLabel =
```
bryophyta, line 33: added to help linting to know whether we're checking all possible cases in the `switch` statement in `handleBadgeClick`

newswires/client/src/SideNav.tsx
```diff
@@ -78,30 +81,6 @@ export const SideNav = () => {
 		[searchHistory],
 	);
 
-	const toggleSupplier = useCallback(
```
bryophyta, line 81: there's some semi-complex logic that we reuse here, so pulling it into the search context instead

newswires/client/src/context/SearchContext.tsx
```diff
@@ -247,32 +250,35 @@ export function SearchContextProvider({ children }: PropsWithChildren) {
...
-			pushConfigState({
-				...currentConfig,
-				query,
-			});
-		} else {
-			pushConfigState({
-				...currentConfig,
-				view: 'feed',
-				query,
+	const handleEnterQuery = useCallback(
```
bryophyta, line 253: this block is just a passby refactor suggested by eslint -- wrapping the function definition in a `useCallback`

newswires/client/.eslintrc.cjs
andrew-nowak: sounds good to me, I'd expect wanting exhaustiveness checking to be a good default, and it's easy to opt out with an eslint-disable directive, or adding a `default` case

sb-dev (approved):

PR #304, opened 2025-05-22
Add ticker routes

sb-dev:
## What does this change?

Add ticker routes:
- `ticker/feed`
- `ticker/item/1`

## Accessibility

### DIFF
--- a/newswires/client/src/urlState.ts
+++ b/newswires/client/src/urlState.ts
@@ -24,6 +24,7 @@ export const defaultConfig: Config = Object.freeze({
 	view: 'home',
 	query: defaultQuery,
 	itemId: undefined,
+	ticker: false,
 });
 
 export function urlToConfig(location: {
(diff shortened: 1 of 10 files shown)


newswires/client/src/urlState.ts
```diff
@@ -70,10 +71,18 @@ export function urlToConfig(location: {
...
-		return { view: 'item', itemId: page.split('/')[1], query };
+	if (page.includes('feed')) {
+		return { view: 'feed', query, ticker: page.includes('ticker') };
+	} else if (
+		page.includes('item/') &&
+		(page.split('/').length === 2 || page.split('/').length === 3)
+	) {
+		return {
+			view: 'item',
+			itemId: page.split('/').pop()!,
```
bryophyta, line 82: small thing, but it would be nice to avoid mutation with `pop` here I think. 

[`Array.at`](https://caniuse.com/?search=array.at) probably doesn't have enough support yet..? But given that all the branches want to match on elements of `page.split`, we could do something like:

```
const pageSegments = page.split("/")

if (pageSegments.includes("feed")) {
    ...
} (...) {
    return {
        ...
        itemId: pageSegments[pageSegments.length - 1]
        ...
    }
}
```

what do you think?

bryophyta (approved):
One small comment but other than that this looks good to me, and works as expected when testing!

I guess we should send out some comms before deploying, because iiuc if a user refreshes an _old_ style ticker after this has deployed, it'll turn into a regular 'feed' view, because it won't have the correct path for a ticker.

newswires/client/src/urlState.ts
sb-dev: yes, pure functions are safer.

PR #316, opened 2025-06-02
Add 'send telemetry as dev' functionality

bryophyta:
Create a new 'sendTelemetryAsDev' flag, sent from the server to the client.
Use this to determine whether to set the 'isDevUser' tag to 'true' or 'false' when sending telemetry from the client.

- [x] Tested on CODE
- [x] Created the new SSM param in the PROD stack

## How to test

- [x] Deploy to CODE
- [x] Open an item in CODE (triggers telemetry) and check that the telemetry event shows up with `_tags.isDevUser: false`
- [x] Add your email address to the SSM param, redeploy, and check that your telemetry now comes in as `isDevUser`.

## Accessibility


newswires/app/controllers/ViteController.scala
```diff
@@ -118,7 +125,7 @@ class ViteController(
 
   private def proxyAsset(
       resource: String,
-      request: Request[AnyContent]
+      request: UserRequest[AnyContent]
```
bryophyta, line 128: I don't _think_ it's a problem to change the expected type here, but I'm not 100%..

newswires/app/controllers/ViteController.scala
```diff
@@ -61,12 +64,16 @@ class ViteController(
...
+        body: String
+    )(implicit request: UserRequest[A]): String = {
       val config =
         views.html.fragments.clientConfig(
           ClientConfig(
             FeatureSwitchProvider.clientSideSwitchStates,
-            stage = configuration.get[String]("stage")
+            stage = configuration.get[String]("stage"),
+            sendTelemetryAsDev =
+              request.user.email.startsWith("pete.faulconbridge@")
```
bryophyta, line 76: hard-coded emails isn't ideal.. is it worth creating an SSM param or something to store emails?

newswires/app/controllers/ViteController.scala
```diff
@@ -47,10 +50,11 @@ class ViteController(
   }
   private val headersToKeep =
     Seq(CACHE_CONTROL, ETAG, DATE, ACCESS_CONTROL_ALLOW_ORIGIN)
+  private val devEmails = configuration.get[String]("devEmails").split(",")
```
bryophyta, line 53: I've created an SSM param: `/PROD/editorial-feeds/newswires/devEmails` (and one for CODE)
not in cloudformation, following the pattern of other runtime SSM params for this app

newswires/app/controllers/ViteController.scala
```diff
@@ -61,12 +65,15 @@ class ViteController(
         .replaceAll("@csrf\\.value", csrf.value)
     }
 
-    def injectClientConfig(body: String): String = {
+    def injectClientConfig[A](
+        body: String
+    )(implicit request: UserRequest[A]): String = {
```
sb-dev, line 70: No need for the following implicit param `implicit request: UserRequest[A]` because it's defined on the parent function.

PR #339, opened 2025-06-16
Revert "step one of gucdk 61.5.0 upgrade"

andrew-nowak:
Now the upgrade has been released, we can remove this temporary code


bryophyta (approved):
great, thanks!

PR #342, opened 2025-06-17
Backend: add presets for individual sports

sb-dev:
## What does this change?

**Path**: /feed?supplier=PA&preset=_**sport-code**_&start=now-2w&end=now

**Example**: `/feed?supplier=PA&preset=olympics&start=now-2w&end=now`

Add individual sports:
- Soccer (code: soccer)
- Golf (code: golf)
- Cricket (code: cricket)
- Rugby Union (code: rugby-union)
- Tennis (code: tennis)
- Cycling (code: cycling)
- Boxing (code: boxing)
- Horse Racing (code: racing)
- F1 (code: f1)
- Rugby League (code: rugby-league)
- Athletics (code: athletics)
- Olympics (code: olympics)

1. Add sport presets for individual sports:
- **Reuters**: Use specific sport topic codes
- **PA**: Use specific sport topic codes when available, fallback to general sport story code and text search
- **AFP**: Use AFP sport code and text search
- **Reuters**: Use specific sport subject codes
- **AP**: Use keywords, and text search when the keyword is not specific enough

2. Add SearchPresets helper to decrease the complexity of search presets

3. Update the full text search allow web-style query syntax:

- Space-separated terms      → AND (e.g.  foo bar      ⇒ foo AND bar)
- OR                                         → OR  (e.g.  foo OR bar)
- –term                                    → NOT (e.g.  foo –bar     ⇒ foo AND NOT bar)
- "exact phrase"                     → exact phrase match
- ( … )                                       → grouping (e.g. foo AND (bar OR baz))

4. Standardise category codes and text searches

5. Replace `OR` with `AND` between keywords and category codes to ensure both filters are applied together, rather than matched separately:

Ex: Soccer Category: apCat:s + Keyword: Soccer

## Accessibility


newswires/app/conf/SearchPresets.scala
```diff
@@ -1,7 +1,31 @@
...
+      supplier: String,
+      categoryCodes: List[String] = Nil,
+      categoryCodesExcl: List[String] = Nil,
+      text: Option[String] = None,
+      searchTerm: Option[SearchTerm] = None,
+      keyword: Option[String] = None,
+      keywordExcl: List[String] = Nil
+  ): SearchParams =
+    SearchParams(
+      text = searchTerm orElse text.map(SearchTerm(_, SearchConfig.Simple)),
```
bryophyta, line 20: if I'm understanding the logic correctly, if both `searchTerm` and `text` are defined on for an instance of `SearchPreset`, then the `text` value will be discarded. Given this, would it be better if we made it so that you can only provide one or the other value when constructing a `SearchPreset`? (either via a function overload or something else?)

newswires/app/conf/SearchTerm.scala
```diff
@@ -0,0 +1,49 @@
...
+ */
+object SimpleSearchQueries {
+  // World preset
+  val REUTERS_NEWS_SCHEDULE = "\"REUTERS NEWS SCHEDULE\""
+
+  // Sport preset
+  val SOCCER = "soccer"
+  val CRICKET = "cricket"
+  val RUGBY_UNION =
+    "(\"rugby league\" \"rugby union\") OR (rugby -\"rugby league\")"
```
bryophyta, line 39: is this first clause correct? `(\"rugby league\" \"rugby union\")`

I'd expect the whole clause for rugby union to be more like: `"\"rugby union\" OR (rugby -\"rugby league\")"`

newswires/test/db/FingerpostWireEntrySpec.scala
```diff
@@ -53,7 +54,7 @@ class FingerpostWireEntrySpec extends AnyFlatSpec with Matchers {
     )
   }
 
-  it should "concatenate keywords and category codes with 'or'" in {
+  it should "concatenate keywords and category codes with 'and'" in {
```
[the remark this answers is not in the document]
bryophyta: I *think* that the answer is that in terms of existing presets it'll only affect ApWorld, and in that case it's probably fine: https://github.com/guardian/newswires/blob/cfffe5894d7d85011b0aabc3c24d84d9dbdf0f92/newswires/app/conf/SearchPresets.scala#L50-L51

But worth documenting I think, and double-checking that there aren't any other side effects?

bryophyta (commented):
this looks good to me, great work! ⚽️🏉🏏🎾

I've just added one more comment about documenting the reason for the change of behaviour ('and' for keywords and categories, rather than 'or')

bryophyta (approved):

newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -275,23 +276,26 @@ object FingerpostWireEntry
       case _ => None
     }
 
-    // grr annoying but broadly I think subjects(/categoryCodes) and keywords are the same "axis" to search on
```
sb-dev, line 278: Removed because keywords can be combined with category codes.
Ex: Soccer `Category: apCat:s` + `Keyword: Soccer`

sb-dev:
Closed due to rebase issue, commits move to https://github.com/guardian/newswires/pull/361

PR #344, opened 2025-06-18
Refactor TS error handling

bryophyta:
There are a number of places where we try to pull a useful message out of an `Error` (or, more accurately, something that's been thrown). We're doing it in slightly different ways in different places.

There's a [blog post about extracting error messages in TypeScript](https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript) which seems to cover most of the possible cases pretty well, so it feels worth adding as a helper function and just using that when needed.

### DIFF
--- a/shared/getErrorMessage.ts
+++ b/shared/getErrorMessage.ts
@@ -19,6 +19,9 @@ function toErrorWithMessage(maybeError: unknown): ErrorWithMessage {
 	if (isErrorWithMessage(maybeError)) return maybeError;
 
 	try {
+		if (typeof maybeError === 'string') {
+			return new Error(maybeError);
+		}
 		return new Error(JSON.stringify(maybeError));
 	} catch {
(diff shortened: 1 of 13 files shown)


shared/getErrorMessage.ts
```diff
@@ -0,0 +1,32 @@
...
+		'message' in error &&
+		typeof (error as Record<string, unknown>).message === 'string'
+	);
+}
+
+function toErrorWithMessage(maybeError: unknown): ErrorWithMessage {
+	if (isErrorWithMessage(maybeError)) return maybeError;
+
+	try {
+		return new Error(JSON.stringify(maybeError));
```
andrew-nowak, line 22: maybe a `if (typeof maybeError === 'string') return new Error(maybeError);`? to avoid unnecessary quoting - not that I can see anything that would cause this

andrew-nowak (approved):

PR #347, opened 2025-06-19
Add 'unique-name' to ingestor input schema

bryophyta:
## What does this change?

* Fingerpost has started sending a `unique-name` field that we are meant to adopt as the unique external id for items coming from them.
* To prepare for this, pass the external id from our pollers as a `unique-name`, too.

## Accessibility


bryophyta:
Actually, it looks like the values we're getting through from Fip might not be unique, so going to put this back in draft while we investigate

PR #355, opened 2025-06-25
remove client dependabot group and exclude react upgrades

andrew-nowak:
## What does this change?

Now we've included the client project into the npm workspace, it's included in the non-cdk-dependencies group, so dependabot is producing duplicate PRs (eg. https://github.com/guardian/newswires/pull/354 and https://github.com/guardian/newswires/pull/352 which both update the client dependencies). In fact the "client" group one is broken, as there is no package-lock in `newswires/client` so dependabot updates only the main `package.json`. Let's remove that group.

Ideally I'd keep the client group and separate it out from the rest, but I can't see a dependabot option to split groups by workspace (since the directory needs to remain the root since that's where the package-lock is)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,6 +17,11 @@ updates:
         update-types:
           - "version-update:semver-major"
           - "version-update:semver-minor"
+      # We can't upgrade to react 19 until eui does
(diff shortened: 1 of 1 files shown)


bryophyta (approved):
👍 👍

PR #360, opened 2025-06-27
Remove total results count from API and UI

bryophyta:
## Accessibility


newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -161,8 +161,8 @@ object FingerpostWireEntry
   }
 
   case class QueryResponse(
-      results: List[FingerpostWireEntry],
```
bryophyta, line 164: I've opted to delete the code downstream, in rendering logic etc., but comment it out at the top of the stream. I'm not sure why but this feels like a reasonable balance..? We should delete it later on though if we aren't going to reinstate for a while.

newswires/client/src/WireItemList.tsx
```diff
@@ -64,17 +58,15 @@ export const WireItemList = ({
 					),
 				)}
 			</ul>
-			{wires.length < totalCount && (
```
bryophyta, line 67: nb. removing the `{wires.length < totalCount && () }` condition means that the 'load more' button will always show, even if there's nothing more to load for the given search (except in the case where there are just no results in the first place)

I think this is reasonable for now -- nothing breaks if the button is pressed when there's nothing more to load, and from what we know of current user behaviour it doesn't seem likely that people are often getting to the end of a list of results like that.

Something to address in a more robust way down the line though, when we know what the longer-term plan is for result counts

paperboyo:
> adding it back in, and optimising other queries where applicable, in follow-ups

Count my vote for all these, FWIW.

PR #362, opened 2025-06-27
UI: add search preset filter options

sb-dev:
## Accessibility


newswires/client/src/SearchSummary.tsx
```diff
@@ -106,22 +109,105 @@ const Summary = ({
...
+			key={value}
+			title={`Filtered by ${label}: ${value}`}
+			iconType="cross"
+			iconSide="right"
+			iconOnClickAriaLabel={`Remove ${label} filter from results`}
+			iconOnClick={() => {
+				handleRemoveBadge(label, value);
+			}}
+			onClick={() => handleBadgeClick()}
+			onClickAriaLabel={''}
```
bryophyta, line 127:
```suggestion
			onClickAriaLabel={undefined}
```

Using `''` seems to render with an empty `aria-label` attribute: 
image

On Chrome at least, the empty attribute seems to be ignored, so the button still has an accessible name, inherited from the text content inside it. So it doesn't actually cause an issue there. But I'm not confident that all browsers will behave the same way, so I'd be worried about the empty aria attribute being interpreted as meaning there's no aria-label for the button.

Although the prop is required on `EuiBadge`, it looks like you can just set it to `undefined`, in which case the `aria-label` attribute is removed altogether. Given that we have an adequate label in the content of the button, I think it makes sense to just have no `aria-label` attribute?

bryophyta (approved):
nice! this looks good to me, and works as expected when testing locally 👍 👍

PR #380, opened 2025-07-08
override jackson versions in project to 2.19.1

andrew-nowak:
## What does this change?

CVE reported in Jackson before 2.15. Try to override to a more modern version, while keeping all the other dependencies in sync.


bryophyta (approved):
this lgtm 👍

PR #385, opened 2025-07-10
Add 'copy headline and URL' button

bryophyta:
Testing out the idea of having a button that will copy the headline and url for an item:

1. Restructure wire item detail header to be more semantic, and separate out the pieces that sit on different lines, so as to enable (2)
2. Add a copy button for the headline and url:
    - 'text/plain': `${headlineText}\n${fullUrl}`
    - 'text/html': `<a href="${fullUrl}">${headlineText}</a>`

image

### DIFF
--- a/newswires/client/src/WireDetail.tsx
+++ b/newswires/client/src/WireDetail.tsx
@@ -332,9 +332,17 @@ function CopyButton({
 				itemId: id.toString(),
 			});
 			const fullUrl = `${window.location.origin}${wireUrl}`;
-			const textToCopy = `${headlineText}\n${fullUrl}`;
 
-			await navigator.clipboard.writeText(textToCopy);
+			await navigator.clipboard.write([
+				new ClipboardItem({
+					'text/plain': new Blob([`${headlineText}\n${fullUrl}`], {
+						type: 'text/plain',
+					}),
+					'text/html': new Blob([`<a href="${fullUrl}">${headlineText}</a>`], {
+						type: 'text/html',
+					}),
+				}),
+			]);
 			setCopied(true);
 			setTimeout(() => setCopied(false), 2000);
 		} catch (err) {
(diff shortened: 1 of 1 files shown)


newswires/client/src/WireDetail.tsx
```diff
@@ -289,6 +315,62 @@ function MetaTable({ wire }: { wire: WireData }) {
...
+			const fullUrl = `${window.location.origin}${wireUrl}`;
+
+			await navigator.clipboard.write([
+				new ClipboardItem({
+					'text/plain': new Blob([`${headlineText}\n${fullUrl}`], {
+						type: 'text/plain',
+					}),
+					'text/html': new Blob([`<a href="${fullUrl}">${headlineText}</a>`], {
+						type: 'text/html',
+					}),
```
andrew-nowak, line 345: this will be vulnerable to some html injection here (likely to be obvious in the rendering of the headline), which could end up in pasting a malicious link instead of one to newswires.
my contrived example is, imagine `headlineText` is `'</a><a href="https://evil.com">A grabbing headline'` - the newswires link ends up with no body, and instead we paste a link to somewhere nefarious.
We can either run the text through `sanitize-html`, or get the browser to do it by creating an `a` element and setting href and body something like this:

```js
const htmlLink = document.createElement('a');
htmlLink.href = fullUrl;
htmlLink.innerText = headlineText;

const htmlLinkBlob = htmlLink.outerHTML;
htmlLink.remove();

await navigator.clipboard.write([
  new ClipboardItem({
    'text/plain': new Blob([`${headlineText}\n${fullUrl}`], {
      type: 'text/plain',
    }),
    'text/html': new Blob([htmlLinkBlob], {
      type: 'text/html',
    }),
  }),
]);
```

andrew-nowak (commented):
mostly lgtm, just one place to be careful about constructing html safely

PR #391, opened 2025-07-22
Fix sinceId query: return 'next page', rather than 'most recent'

bryophyta:
## What does this change?

The main change here is to distinguish between "next page" vs "most recent" searches in the `QueryParams` class, and to change the ordering of results in the SQL query depending on the search type.

### Paging bug

The feed in the Newswires UI is constructed as follows (when auto-refresh is on):

- On initial search, fetch the `n` most recent items matching the search terms `S`.
- Periodically make requests with the same terms `S`, but with an added parameter of `sinceId` that's set to the most recent id in the client's store.
- If there are new items, merge them with the existing feed results.

This fixes a bug in the existing code, whereby if you search using a `sinceId` parameter (as we do when fetching updates to the feed in the client app) you'll get the `n` most recent items. In the case where there have been more than `n` new items since the item with `sinceId`, the feed will contain a gap.

## Accessibility

### DIFF
--- a/newswires/app/db/QueryParams.scala
+++ b/newswires/app/db/QueryParams.scala
@@ -2,11 +2,18 @@ package db
 
 import conf.SearchTerm
 
+sealed trait UpdateType {
+  val sinceId: Int
+}
+
+case class NextPage(sinceId: Int) extends UpdateType
+case class MostRecent(sinceId: Int) extends UpdateType
+
 case class QueryParams(
     searchParams: SearchParams,
     savedSearchParamList: List[SearchParams],
     maybeSearchTerm: Option[SearchTerm],
     maybeBeforeId: Option[Int],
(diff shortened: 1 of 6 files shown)


newswires/app/db/QueryParams.scala
```diff
@@ -0,0 +1,19 @@
+package db
+
+import conf.SearchTerm
+
+sealed trait UpdateType {
+  val sinceId: Int
+}
+
+case class NextPage(sinceId: Int) extends UpdateType
+case class MostRecent(sinceId: Int) extends UpdateType
```
bryophyta, line 10: `MostRecent` is not currently used, but I am planning to use it in an upcoming PR. Can remove it here though if we aren't happy with adding the abstraction prematurely.

PR #400, opened 2025-07-28
Add a queueing lambda that just passes messages straight through for now

bryophyta:
## What does this change?

Adds a new lambda -- the `Fingerpost queueing lambda` -- in between the Fingerpost SNS topic (and associated SQS queue) and the Ingestion lambda. In this initial implementation, it just creates a new item on the SQS 'Source queue' (the one that's used by our custom pollers to send work to the Ingestion lambda) without doing anything else to the content.

Initial implementation is just temporary, to prove that the right wiring is in place. In subsequent PRs the new Queueing lambda will be responsible for putting incoming content into a bucket, that the Ingestion lambda will then read from.

## How to test

- [x] Deploy to CODE and check that we're still successfully ingesting items from:
    - [x] Fingerpost
    - [x] Our own AP poller
    - [x] Our own Reuters poller

After subsequent work is done, the ingestion pipeline will:

1. have a cleaner separation of concerns between the different steps
2. be more robust (i.e. able to cope with content that's too big for an SNS or SQS event, because it'll use S3 as an intermediate storage step)
3. be easier to re-run, in the eventuality that we lose data from the database, because the proximate input for ingestion will just be an event with an S3 key in it, rather than with the whole content in it

## Accessibility

### DIFF
--- a/fingerpost-queueing-lambda/README.md
+++ b/fingerpost-queueing-lambda/README.md
@@ -1,3 +1,3 @@
-# Cleanup Lambda
+# Fingerpost Queueing Lambda
 
-A simple Lambda function that runs on a schedule, deleting old records from the database.
+A Lambda function that's invoked by incoming items on the Fingerpost SNS topic, and is responsible for adding them to the 'source queue' for the Ingestion Lambda.


poller-lambdas/src/aws.ts
```diff
@@ -1,28 +1,8 @@
...
-		send: (command: SendMessageCommand) => {
-			const currentQueue: SendMessageCommand[] =
-				queueData[command.input.QueueUrl!] ?? [];
-			queueData[command.input.QueueUrl!] = [...currentQueue, command];
-			return Promise.resolve();
-		},
-	};
-};
-
-export const sqs = isRunningLocally
```
bryophyta, line 23: `sqs` client construction is being moved to `../../shared` because the queueing lambda needs to use it too now

shared/config.ts
```diff
@@ -6,24 +8,19 @@
 export const isRunningLocally =
 	!process.env.LAMBDA_TASK_ROOT && !process.env.AWS_EXECUTION_ENV;
 
-export const DATABASE_NAME: string = isRunningLocally
```
bryophyta, line 9: moving this to the `rds` module so that projects which don't use the db (and so don't have the relevant environmental variables) can import values from this config file

PR #401, opened 2025-07-29
fix setup overwrite_config to fully overwrite

andrew-nowak:
## What does this change?

When using the "overwrite" option in the setup script, it's currently just appending the values to the existing config file. This should work as the config lib reads top-to-bottom, so later values will replace earlier ones in memory, but it's still nice to clean up as we go.

## How to test

Use the overwrite option a few times, you should end with a config file that's a little bit neater.

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -22,6 +22,9 @@ config_pairs=(
 )
 
 overwrite_config() {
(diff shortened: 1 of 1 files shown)


bryophyta (approved):
🙏 🙏

PR #404, opened 2025-07-30
Refactor S3 utils

bryophyta:
## What does this change?

- Refactor S3 utils so that we export get and put functions, rather than the S3 client
- Move the utils to the shared js directory, instead of the ingestion-lambda project

This is preparation for work to make the ingestion lambda read its contents from S3, when other projects (the fingerpost-queueing-lambda and the poller-lambdas) will need to interact with S3 too.

## Accessibility

### DIFF
--- a/shared/s3.ts
+++ b/shared/s3.ts
@@ -6,6 +6,7 @@ import {
 } from '@aws-sdk/client-s3';
 import { getFromEnv, isRunningLocally } from './config';
 import { getErrorMessage } from './getErrorMessage';
+import { createLogger } from './lambda-logging';
 import type { OperationResult } from './types';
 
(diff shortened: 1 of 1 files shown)


shared/s3.ts
```diff
@@ -0,0 +1,90 @@
...
+const s3Client = new S3Client(awsOptions);
+
+export async function getFromS3({
+	bucketName,
+	key,
+}: {
+	bucketName: string;
+	key: string;
+}): Promise<OperationResult<{ body: string }>> {
+	console.log({
```
andrew-nowak, line 33: should this be using the logger util?

andrew-nowak (approved):

PR #405, opened 2025-07-31
Strip br tags in table

lindseydew:
## What does this change?

This updates our html parsing of table data to remove any <br> tags. 

This is because when they are left in the table, on the front end when when render the table this function: https://github.com/guardian/newswires/blob/d4022b9da707ce7372e88c9d1a0ec7c36653aba3/newswires/client/src/WireDetail.tsx#L503 will transform all the <br> tags within the table to the top of the table. Which is why we end up with a large amount of white space at the moment. 

## How to test

I have only unit tested this. Would be good to get advice as to whether this is worth regressions testing too 

## Accessibility

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -22,6 +22,9 @@ config_pairs=(
 )
 
 overwrite_config() {
+  # first, delete existing config contents
+  echo > ~/.gu/newswires.conf
+  # then write new contents
   for pair in "${config_pairs[@]}"; do
       echo "$pair" >> ~/.gu/newswires.conf
   done


bryophyta (approved):
This looks good to me! Unit test looks like it should cover the right behaviour, but happy to chat about testing in CODE if it'd be useful

lindseydew:
> This looks good to me! Unit test looks like it should cover the right behaviour, but happy to chat about testing in CODE if it'd be useful

Ok, maybe we could do a code regression test together later on something with a bit more of a significant change 👍

PR #408, opened 2025-07-31
Ingestion goes via intermediate S3 storage step

bryophyta:
Both the fingerpost-queueing-lambda and the poller-lambdas write content to S3 and then queue work for the ingestion-lambda, which reads the content from S3.

This will allow us to handle content that is too big to fit in an SQS message, and also make it easier to re-ingest content if we need to.

It also makes way for the sport.copy SES integration.

## Accessibility


lindseydew (approved):
LGTM 👍

bryophyta:
This has been rebased a couple of times so just did one last end-to-end test on CODE; all looks good so going to go ahead and merge

PR #413, opened 2025-08-05
SES message processing

bryophyta:
## What does this change?

- Add a code path to the ingestion-lambda which will ingest SES messages in largely the same way as SQS messages.
- Move the email verification function from the email-filter-lambda project to the shared project, so that we can apply verification in the ingestion lambda too.
- Add a `/copy.json` endpoint to the main app, to allow us to see processed email messages for debugging. (Proper UI for email messages will be added later.)

## How to test

- [x] Deploy to CODE, send a test email, check whether it shows up at `/copy.json` (unless flagged as spam etc.)

## Accessibility


ingestion-lambda/src/handler.ts
```diff
@@ -1,17 +1,49 @@
...
+import { processEmailContent } from './processEmailContent';
 
-function processMessage(
+function processSESRecord(record: SESEventRecord): OperationResult<{
+	externalId: string;
+	objectKey: string;
+}> {
+	const { ses } = record;
+
+	const { hasFailures, failedChecks } = findVerificationFailures(ses.receipt);
```
bryophyta, line 25: adding the verification failure checks here because the email filtering lambda hasn't been switched on properly yet, but when it has we can delegate to it if we want to

ingestion-lambda/src/processEmailContent.test.ts
```diff
@@ -0,0 +1,62 @@
...
+		const emailData = sampleMimeEmailData();
+		const { from, subject, text, date } = await parseEmail(emailData);
+		expect(from).toBe('"Correspondent A" <correspondent.a@example.com>');
+		expect(subject).toBe('Test subject line');
+		expect(text).toContain('This is some test copy: hello world');
+		expect(date).toBe('Wed, 16 Jul 2025 08:38:29 GMT');
+	});
+});
+
+export function sampleMimeEmailData() {
```
lindseydew, line 14: Since this is returning static data I think this can be a const here?

ingestion-lambda/src/handler.test.ts
```diff
@@ -267,21 +274,78 @@ describe('handler.main', () => {
...
+		mockSql.mockResolvedValue([{ id: 1 }]);
+
+		mockInitialiseDbConnection.mockResolvedValue({
+			sql: mockSql as unknown as postgres.Sql,
+			closeDbConnection: jest.fn(),
+		});
+
+		const sesEvent = {
+			Records: [
+				{
```
lindseydew, line 293: Small point, but might be nice to move the creation of this object and call it something to show it's successful, and then you can make a another one that just varies the spam verdict

ingestion-lambda/src/handler.ts
```diff
@@ -31,45 +63,64 @@ function processMessage(
...
+		return processSESRecord(record);
+	} else {
+		return processSQSRecord(record);
+	}
+}
+
+function isSESRecord(
+	record: SESEventRecord | SQSRecord,
+): record is SESEventRecord {
+	return 'ses' in record;
```
lindseydew, line 79: Is this fully robust? Could an sqs record that has the string ses pass this check? Might be unlikely but could be worth seeing if there is a specific field to key from.

lindseydew (commented):
Looking really good. Just a few small comments

newswires/app/controllers/QueryController.scala
```diff
@@ -48,9 +63,10 @@ class QueryController(
     val maybePreset =
       request.getQueryString("preset").flatMap(SearchPresets.get)
 
-    val suppliersToExcludeByDefault =
-      if (featureSwitchProvider.ShowGuSuppliers.isOn()) Nil
-      else List("GuReuters", "GuAP")
+    val suppliersToExcludeByDefault: List[String] = List(
+      "UNAUTHED_EMAIL_FEED"
+    ) ++ (if (featureSwitchProvider.ShowGuSuppliers.isOn()) List.empty
+          else List("GuReuters", "GuAP"))
```
lindseydew, line 70: I'd perhaps express this as: 

```
val suppliersToExcludeByDefault =
      if (featureSwitchProvider.ShowGuSuppliers.isOn()) List("UNAUTHED_EMAIL_FEED")
      else List("GuReuters", "GuAP", "UNAUTHED_EMAIL_FEED")
``` 

Even though it duplicates the email supplier, it feels a bit easier not to have to do string concat mentally. 


This this could also be good to perhaps move out of the main function and write a unit test for (more for documenting business logic than anything else)

lindseydew: It would probably also be good to unit test the resolution of `suppliersExcl`. 

This expression: 

```
request.queryString
        .get("supplierExcl")
        .map(_.toList)
        .getOrElse(Nil) ++ suppliersToExcludeByDefault.filterNot(
        suppliers.contains
      )
```

Because there are a few variations on this depending on whether the switch is on, or what is present in the suppliers query. 

(Though the 
```
request.queryString
        .get("supplierExcl")
        .map(_.toList) 
```
could be computed else where so you don't have to worry about having a request object in the unit test. 

Let me know if it doesn't make sense

ingestion-lambda/src/processEmailContent.test.ts
bryophyta: good call -- my reason for making it a function was just so that it could go at the bottom of the file while I was developing, but , but I agree that a `const` is more expressive of the intent so I've updated it now, exporting from its own file as it's actually consumed in two places now 👍👍

ingestion-lambda/src/handler.ts
lindseydew: Ok cool 👍

PR #415, opened 2025-08-05
Need to vary the composer host to match the stage!

andrew-nowak:
## What does this change?

Oversight in #382  - I forgot to come back and fix where I'd hardcoded composer's CODE domain on insertion to the new database 🤦 

Come back and do it quickly, hopefully before anyone notices.

## How to test

Hard to see the fix on CODE, since it already was storing the right URL... At most checking that it hasn't caused a regression should be enough.


paperboyo:
I never noticed. Honest.

PR #422, opened 2025-08-07
Telemetry for 'fetched results', including round-trip duration

bryophyta:
## Accessibility


newswires/client/src/context/SearchContext.test.tsx
```diff
@@ -84,7 +84,6 @@ describe('SearchContext', () => {
 			contextRef.current?.handleEnterQuery(q);
 		});
```
lindseydew, line 86: These being removed because the line below effectively does the same check?

lindseydew (approved):
LGTM 👍

newswires/client/src/context/SearchContext.test.tsx
bryophyta: These assertions started failing because we're sending two telemetry events now instead of 1. But I wasn't sure whether we want to update the test case, or to remove assertions about when/how telemetry is called. So I opted to avoid that problem and just delete this assertion for now 😅 

I don't really know which direction to go though; very open to suggestions!

PR #429, opened 2025-08-11
do not exclude MCC:SPO-tagged Reuters World stories from World section

andrew-nowak:
## What does this change?

We're missing a small handful of stories where they have been tagged with REUTERS:WORLD, but then we exclude them as they also contains MCC:SPO. Stories in World may have sports themes, so we should allow them to appear in the feed.


paperboyo:
I _think_ it makes sense. 31 stories (of 2,745) over a span of a week extra in World sounds good?

PR #432, opened 2025-08-11
Ld/format s3 key

lindseydew:
## Accessibility


fingerpost-queueing-lambda/src/handler.ts
```diff
@@ -13,7 +14,7 @@ export const main = async (event: SQSEvent): Promise<SQSBatchResponse> => {
 				const externalId = messageAttributes['Message-Id']?.stringValue;
 				const hasExternalId = externalId && externalId.trim().length > 0;
 				const objectKey = hasExternalId
-					? `${externalId}.json`
+					? formatS3Key(externalId)
```
bryophyta, line 17: This looks good to me. Should the formatting be fingerpost-specific or apply to all items going via S3? If the latter, should we add this to the poller lambdas too?

shared/formatS3key.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    });
+    it('should replace spaces with a hyphen', () => {
+        const externalId = 'test external id';
+        const expectedKey = 'test-external-id.json';
+        expect(formatS3Key(externalId)).toBe(expectedKey);
+    });
+    it('should replace forward slashes with a hyphen', () => {
+        const externalId = 'test/external/id';
+        const expectedKey = 'test-external-id.json';
+        expect(formatS3Key(externalId)).toBe(expectedKey);
```
andrew-nowak, line 17: Does this run the risk that some stories might overwrite each other in S3? if we had 3 files arrive with each of these external ids, then we'd be left with only the most recent, even though they've each got different ids. (Hopefully actually we still have all 3, but only as versions of object with the same key). 
Do we have a concrete problem with the spaced or slashed keys? Maybe we run the id through a url encoder instead?

PR #433, opened 2025-08-12
extra ignored test which can output the sql query for a set of scala …

andrew-nowak:
…params

## Accessibility

### DIFF
--- a/newswires/test/db/FingerpostWireEntrySpec.scala
+++ b/newswires/test/db/FingerpostWireEntrySpec.scala
@@ -1,6 +1,6 @@
 package db
 
-import conf.{SearchField, SearchTerm}
+import conf.{SearchField, SearchPresets, SearchTerm}
 import io.circe.parser.decode
 import helpers.SqlSnippetMatcher.matchSqlSnippet
 import helpers.models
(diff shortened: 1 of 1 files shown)


newswires/test/db/FingerpostWireEntrySpec.scala
```diff
@@ -320,4 +321,53 @@ class FingerpostWireEntrySpec extends AnyFlatSpec with Matchers with models {
...
+    // This is a textbook example of how to get hit by a SQL injection attack - this should be used
+    // for troubleshooting only, and if any part of the query comes from an untrusted source,
+    // you really need to scan the output to check that no part is attempting to perform malicious actions.
+    val fmted = query.parameters.foldLeft(
+      query.statement.replaceAll("\\?\\?", "?")
+    )((statement, param) => {
+      val matcher = "\\?(?![|?])"
+      param match {
+        case p: Int    => statement.replaceFirst(matcher, s"$p ")
+        case p: String => statement.replaceFirst(matcher, s"'$p' ")
```
bryophyta, line 364: just spotted that this will fail if the param is the name of a column, because that doesn't want to be placed inside `'`s.

we are currently doing that in some queries (I think here? https://github.com/guardian/newswires/blob/d00ab66faa21d9f8a046d00154673f71cdd6d490/newswires/app/db/FingerpostWireEntry.scala#L288-L293).

not sure if it's worth trying to account for this edge case in the code, but thought i'd flag it as something to be aware of for future travellers :)

PR #447, opened 2025-08-26
Add 'dotcopy' endpoint

bryophyta:
## What does this change?

- Add a new `api/dotcopy` route, which serves dotcopy items, and allows search/filtering on the following fields: `(q: Option[String], start: Option[String], end: Option[String], beforeId: Option[Int], sinceId: Option[Int])`
- As part of adding the dotcopy handler, I've added a `baseWhereClause` option to the where clause builder. This allows us to explicitly specify whether to include or exclude the email feed from results, close to the place where it matters (i.e. in the `query` and `dotcopy` functions). This allows us to remove the `QueryController.effectiveSuppliersToExclude` method, which (as @lindseydew pointed out) was hard to reason about.
- Refactor how we generate highlight clauses -- there are three separate places where we construct almost-identical clauses now, and there is some logic in there, so it felt like it made sense to abstract it.

## How to test

- [x] Check `api/dotcopy` and make sure emails are showing up, and that the filters work as expected.
- [x] Check that we still aren't showing emails in the main feed (e.g. search for 'Faulconbridge' today in the main feed and make sure my email from earlier doesn't show up)
- [x] Verify that highlights still work as expected

## Accessibility


newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -124,11 +125,29 @@ object FingerpostWireEntry
...
   private[db] def buildSingleGetQuery(
       id: Int,
-      maybeFreeTextQuery: Option[String]
+      maybeFreeTextQuery: Option[SearchTerm]
   ): SQLSyntax = {
-    val highlightsClause = maybeFreeTextQuery match {
+    val highlightsClause =
+      buildHighlightsClause(maybeFreeTextQuery, highlightAll = true)
+
+    maybeFreeTextQuery match {
```
andrew-nowak, line 150: we can delete these lines now right? since they've been extracted above?

PR #454, opened 2025-09-02
Tweaks for displaying dataformatted items in feed list

bryophyta:
## What does this change?

- Add a 'table' icon to the feed item for stories with data formatting
- Fix the text wrapping and truncation for subheadings. 
  - Before we were replacing certain html tags with non-breaking spaces, but this mean that the truncation didn't work and text was overflowing its container sometimes.
  - I've looked to see if there are any unintended consequences of this change for rendering other items but I haven't been able to spot any yet

### Regular items are unchanged 

image

### Items with data formatting have a new icon

image

### Text truncation on subheadings is fixed

(See the third item down on both -- on the left the text is breaking out of the list)

image

## Accessibility

### DIFF
--- a/newswires/client/src/WireItemList.tsx
+++ b/newswires/client/src/WireItemList.tsx
@@ -1,5 +1,6 @@
 import {
 	EuiButton,
+	EuiIcon,
 	EuiScreenReaderOnly,
 	EuiText,
 	EuiTextBlockTruncate,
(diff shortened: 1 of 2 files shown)


newswires/client/src/WireItemList.tsx
```diff
@@ -146,7 +147,7 @@ function MaybeSecondaryCardContent({
 		return <p>{subhead}</p>;
 	}
 	const maybeBodyTextPreview = bodyText
-		? sanitizeHtml(bodyText.replace(/(<br \/>|<\/p>)/g, '&nbsp;'), {
+		? sanitizeHtml(bodyText.replace(/(<br \/>|<\/p>|<\/td>)/g, ' '), {
```
lindseydew, line 150: Curios about the nbsp change to a normal space here?

lindseydew (approved):
LGTM Nice one 👍

newswires/client/src/WireItemList.tsx
bryophyta: I can't find a good reference for this, but my understanding is that the usual line-wrapping algorithm won't break a line on an `&nbsp;` in the way it does for regular spaces. I think the issue here could also be addressed with CSS (something like `word-break: break-all;`) but this felt more straightforward, and less likely to affect other kinds of subheadings?

PR #455, opened 2025-09-02
Deduplicate slug in feed for items with data formatting

bryophyta:
## What does this change?

We've started getting the headlines through from Fip for items with data formatting, but the headline includes the slug. So we shouldn't add the slug to the headline in the feed view, for items with data formatting.

### Before

image

### After

image

## Accessibility

### DIFF
--- a/newswires/client/src/Item.stories.tsx
+++ b/newswires/client/src/Item.stories.tsx
@@ -37,6 +37,7 @@ const sampleItemData: WireData = {
 	ingestedAt: '2025-02-26T09:58:22.000Z',
 	categoryCodes: ['C:US', 'C:CA'],
 	isFromRefresh: false,
+	hasDataFormatting: false,
 };
 
 const meta = {
--- a/newswires/client/src/Item.stories.tsx
+++ b/newswires/client/src/Item.stories.tsx
@@ -2,6 +2,7 @@ import { EuiProvider } from '@elastic/eui';
 import type { Meta, StoryObj } from '@storybook/react-vite';
 import { SearchContextProvider } from './context/SearchContext';
 import { TelemetryContextProvider } from './context/TelemetryContext';
+import { UserSettingsContextProvider } from './context/UserSettingsContext';
 import { setUpIcons } from './icons';
 import { Item } from './Item';
 import type { WireData } from './sharedTypes';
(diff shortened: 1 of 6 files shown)


newswires/client/src/Item.stories.tsx
```diff
@@ -50,11 +52,13 @@ const meta = {
 		(Story) => (
 			<EuiProvider colorMode="light">
 				<TelemetryContextProvider sendTelemetryEvent={console.log}>
-					<SearchContextProvider>
-						<div style={{ maxWidth: '800px', margin: '0 auto' }}>
-							<Story />
-						</div>
-					</SearchContextProvider>
+					<UserSettingsContextProvider>
```
lindseydew, line 55: I might be missing something here, but I'm curious as to why this provider has been introduced and not used?

newswires/client/src/context/transformQueryResponse.test.ts
```diff
@@ -37,7 +38,67 @@ describe('transformWireItemQueryResult', () => {
 			...input,
 			supplier: UNKNOWN_SUPPLIER,
 			ingestedAt: '2025-01-02T00:00:00+00:00',
+			hasDataFormatting: false,
 		};
 		expect(transformWireItemQueryResult(input)).toEqual(expectedOutput);
 	});
+
```
lindseydew, line 45: 😍

newswires/client/src/WireItemList.tsx
```diff
@@ -75,17 +84,26 @@ export const WireItemList = ({
...
-			: '';
+	const hasNonEmptySlug = slug && slug.length > 0;
 
 	if (headline && headline.length > 0) {
+		/**
+		 * AAP and PA stories have useful slugs. But stories with 'data formatting' have
+		 * their slugs added to their headlines when we get them, so we don't need to add them again.
+		 */
+		const prefix =
+			(supplier === 'AAP' || supplier === 'PA') &&
```
lindseydew, line 97: I feel like these conditions could be abstracted out to perhaps a `showPrefix` boolean to make this ternary a bit simpler to parse

lindseydew (commented):
Nice looking good, just a couple of small comments

newswires/client/src/Item.stories.tsx
bryophyta: We wrote the the `useUserSettings` hook so that it throws an error at runtime if it's called outside of a `UserSettingsContextProvider` instance. In this case, the `Item` component has a child that uses the `useUserSettings` hook: https://github.com/guardian/newswires/blob/a9702cd7d0ca1092d2fcfa9e787565d98e4f502a/newswires/client/src/ToolsConnection.tsx#L181

I think the broader pattern comes from this post: https://kentcdodds.com/blog/how-to-use-react-context-effectively#typescript

newswires/client/src/WireItemList.tsx
bryophyta: good call, thanks! I've done that now 👍

PR #462, opened 2025-09-10
Search with modifiers for Category Codes

paperboyo:
_Co-authored by: copilot_

## What does this change?

À la Grid, this adds ability to exclude a certain category (add `categoryCodeExcl` filter) when holding Alt key over a Category Codes filter. It also changes the icon to `minusInCircle` when you hold Alt, but not for `check` icons (this isn’t strictly necessary, but I thought it nice and Chat said performance won’t be affected). Similarly to clicking plus, clicking minus adds and then removes `categoryCodeExcl` filter (toggle). We have not handled the case of clicking a `check` icon while holding Alt, because.

## How to test

Find a wire story with at least one Category Code, hold Alt observe the icon change, click and see the `categoryCodeExcl` added to query. Verify no other icons but plus icons only in Category Codes change when you hold Alt. Observe that we do not indicate a fact that certain category has been excluded, because a) it would be mad and b) it would ever only apply to the very wire one excluded a category from, because any subsequent ones from this query wouldn’t have that very category.
 
Users (yeah, right…) can focus their searches interactively, discovering the makeup of the corpus easier as-they-go without resorting to copy/pasting bits of URLs… 

## Have we considered potential risks?

We absolutely have not. We asked Mr. Chat about performance, but we blindly believed it. Plus, we haven’t asked Mr. Chat about the risk it poses itself. 

## Images

hover

## Accessibility

-   [x] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/newswires/client/src/WireDetail.tsx
+++ b/newswires/client/src/WireDetail.tsx
@@ -115,11 +115,19 @@ function CategoryCodeTable({ categoryCodes }: { categoryCodes: string[] }) {
 		const handleKeyUp = (e: KeyboardEvent) => {
 			if (!e.altKey) setIsAltPressed(false);
 		};
+		// Reset state when window loses focus (e.g., Alt+Tab)
+		const handleBlur = () => {
+			setIsAltPressed(false);
+		};
+
 		window.addEventListener('keydown', handleKeyDown);
 		window.addEventListener('keyup', handleKeyUp);
+		window.addEventListener('blur', handleBlur);
+
 		return () => {
 			window.removeEventListener('keydown', handleKeyDown);
 			window.removeEventListener('keyup', handleKeyUp);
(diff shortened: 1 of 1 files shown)


bryophyta (approved):
At first I was thinking that the key event handlers should be managed with the global shortcut key logic, but actually because the alt key logic a) only affects presentation within this component, and b) doesn't compete with global shortcut keys, I think it makes sense to have it in this component.

I've added one extra case to the useEffect callback, to reset the altkey state when the user navigates away from the window. (Otherwise, if `alt` is held down when the user goes away, the `keyup` event won't be caught by the window and so it'll still think the alt key's being held down when the user comes back)

PR #464, opened 2025-09-11
Ingestion time

lindseydew:
## What does this change?
This reads the `lastModified` time from the s3 file and uses this to persist the `ingested_time`, so that if we need to rerun the queue at a separate time the time persisted in the database can be tied to the document rather than when we ran the script. 

## How to test
Tested rerunning the dead letter queue and seeing documents be persisted in order. 

## Accessibility

### DIFF
--- a/cdk/lib/__snapshots__/whole-stack.test.ts.snap
+++ b/cdk/lib/__snapshots__/whole-stack.test.ts.snap
@@ -60,7 +60,7 @@ exports[`createStacks should create WiresFeeds and Newswires stacks with correct
     },
   },
   "Resources": {
-    "FingerpostDeadLetterQueueAlarm577676D8": {
+    "fingerpostDeadLetterQueueAlarm105C0EED": {
       "Properties": {
         "ActionsEnabled": false,
         "AlarmActions": [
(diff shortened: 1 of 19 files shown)


ingestion-lambda/src/db.ts
```diff
@@ -11,21 +11,24 @@ export async function putItemToDb({
...
 }): Promise<OperationResult<{ didCreateNewItem: boolean }>> {
 	const { content, supplier, categoryCodes } = processedObject;
+	const ingestedAt = lastModified ?? new Date();
 	try {
 		const result = await sql`
 		INSERT INTO ${sql(DATABASE_TABLE_NAME)}
-			(external_id, supplier, content, category_codes, s3_key)
-		VALUES (${externalId}, ${supplier}, ${content as never}, ${categoryCodes}, ${s3Key}) ON CONFLICT (external_id) DO NOTHING
+			(external_id, supplier, content, category_codes, s3_key, ingested_at)
+		VALUES (${externalId}, ${supplier}, ${content as never}, ${categoryCodes}, ${s3Key}, ${ingestedAt}) ON CONFLICT (external_id) DO NOTHING
```
bryophyta, line 31: have we verified that this will use the same format and timezone etc. as the current `ingested_at` column does?

PR #465, opened 2025-09-14
Further overhaul of Sport presets

paperboyo:
This builds on @lindseydew’s work in `ld/add-sports-presets` (which, among other things, allows for slug searches, yay!)
  
## What does this change?

- Fixes sort order of Rugby presets in the dropdown
- Renames F1 to Motor Racing (and broadens its scope to all car racing), but without motorbikes, so maybe should be called Car Racing? Or include motorbikes?
- Adds Cycling Results presets to declutter cycling stories
- I went through all Sport presets, tightened them and separated results as best I could
- To improve it, we need two things I can’t see us having currently:
   - ability to specify a preset by searching in multiple fields at once (eg. AP Soccer should be Slug: `fbl` and Headline: `-table -results`)
   - ability to exclude subsuppliers (`sourceFeed` field) (eg. some PA Rugby Results arrive as `pa:SSS`, but should be excluded by`sourceFeed:-"PA PA SPORT DATA"`)
- We should probably rename `All Data Formats` to `All Sport Tables` (but there is a broaden question of making TYPES of stories more evident/accessible)

## How to test

Best to compare with naked `ld/add-sports-presets`, I suppose. I would give details of what this PR does, but unsure if current CODE contains naked `ld/add-sports-presets`.

This seems ready(ish) for human consumption. 

## Accessibility

### DIFF
--- a/newswires/app/conf/SearchPresets.scala
+++ b/newswires/app/conf/SearchPresets.scala
@@ -90,7 +90,8 @@ object SearchPresets {
     case "tennis"               => Some(Tennis)
     case "tennis-results"       => Some(TennisResults)
     case "cycling"              => Some(Cycling)
-    case "f1"                   => Some(F1)
+    case "cycling-results"      => Some(CyclingResults)
+    case "motor-racing"         => Some(MotorRacing)
     case "golf"                 => Some(Golf)
     case "golf-results"         => Some(GolfResults)
     case "boxing"               => Some(Boxing)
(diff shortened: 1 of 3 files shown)


newswires/app/conf/SearchPresets.scala
```diff
@@ -254,80 +312,146 @@ object SearchPresets {
...
+    SearchPreset.fromSearchTerm(
+      REUTERS,
+      searchTerm = SearchTerm.Simple("-(OPTA)", SearchField.BodyText),
+      CategoryCodes.RugbyUnion.REUTERS
+    ),
+    SearchPreset.fromSearchTerm(
+      PA,
+      searchTerm = SearchTerm.Simple("RUGBYU", Slug),
+      CategoryCodes.Sport.PA
+    ), // make excluding "sourceFeed": "PA PA SPORT DATA"
```
bryophyta, line 379: bookmarking this to be able to link to from a ticket

paperboyo:
Closing in favour of https://github.com/guardian/newswires/pull/527.
Will dig out actual searches from here and resubmit once the proper PR is in.

PR #472, opened 2025-09-19
Remove sideNav state from SearchContext; pass as prop from App instead

bryophyta:
## What does this change?

This is a preparatory step towards refactoring SearchContext. Side nav state isn't connected directly to any of the search state management, so it's good to manage it separately. I thought about creating a new context provider for it, but prop-drilling works fine at the moment, and it'd be easy to switch to a new context later if we want.

### DIFF
--- a/newswires/client/src/SearchSummary.tsx
+++ b/newswires/client/src/SearchSummary.tsx
@@ -291,12 +291,15 @@ const Summary = ({
 	);
 };
 
-export const SearchSummary = () => {
+export const SearchSummary = ({
+	setSideNavIsOpen,
+}: {
+	setSideNavIsOpen: (isOpen: boolean) => void;
+}) => {
 	const {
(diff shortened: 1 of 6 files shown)


newswires/client/src/SearchSummary.tsx
```diff
@@ -291,12 +291,15 @@ const Summary = ({
 	);
 };
 
-export const SearchSummary = () => {
+export const SearchSummary = ({
```
bryophyta, line 294: The use of `setSideNavIsOpen` in the search summary here (for ticker layouts) is the reason why we need to raise the side nav state up. It would be nice to restructure the layout so that we don't need this, but that would be a bigger change.

lindseydew (approved):
Code looks good to me 👍

I wasn't totally clear on the rationale behind it, but not a blocker for merging

PR #474, opened 2025-09-19
Split main feed/item layout into separate components

bryophyta:
## What does this change?

There's quite a lot of nested/interleaved logic to handle the differences in layout between the default view and the ticker view. This PR creates two separate 'layout' components which aim to encapsulate the distinct layouts of each view, so that we can just have a single decision point about which view to render in the App component.

Functionality and appearance should all be the same as before, except that I've made sure the telemetry pixel is added to both views by including it in the shared App parent.

### DIFF
--- a/newswires/client/src/App.tsx
+++ b/newswires/client/src/App.tsx
@@ -169,6 +169,7 @@ export function App() {
 							synchronization will resume once connectivity is restored."
 							color="warning"
 							iconType="warning"
+							size="s"
 						/>
 					)}
 					{isRestricted(config.query.dateRange?.end) &&
@@ -179,6 +180,7 @@ export function App() {
 								filter to see the latest data."
 								color="warning"
 								iconType="warning"
+								size="s"
 							/>
 						)}
 					{!!config.ticker && (
--- a/newswires/client/src/App.tsx
+++ b/newswires/client/src/App.tsx
@@ -1,15 +1,8 @@
 import {
 	EuiButton,
 	EuiButtonEmpty,
-	EuiButtonIcon,
 	EuiCallOut,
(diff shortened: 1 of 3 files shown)


newswires/client/src/App.tsx
```diff
@@ -107,15 +103,18 @@ export function App() {
 		} else {
 			document.title = 'Newswires';
 		}
-	}, [isPoppedOut, config.query]);
+	}, [isTickerView, config.query]);
 
-	const largeMinBreakpoint = useEuiMinBreakpoint('l');
-	const largeMaxBreakpoint = useEuiMaxBreakpoint('l');
+	const errorPromptComponent = useMemo(() => {
```
lindseydew, line 108: Just checking my understanding here. This component here will store an error based on whether there has been an error. If the status changes, this component would remain populated? 

It looks like there is an explicit check for the `status` where it is used further down, so I guess it would be fine for this component to remain populated even after a state change. Just checking the behaviour

newswires/client/src/App.tsx
```diff
@@ -174,215 +173,26 @@ export function App() {
...
-										Clear
-									</EuiButton>,
-								]}
-								body={<p>Sorry, failed to load because of {state.error}</p>}
-								hasBorder={true}
-							/>
-						)}
-					</div>
+					{!!config.ticker && (
+						<TickerLayout errorPromptComponent={errorPromptComponent} />
```
lindseydew, line 191: I'm curious why the component is being passed in as an explicit prop, rather than making the error component a child of the parent?
```
<TickerLayout>
 { status === 'error' && <ErrorComponent/> }
</TickerLayout>
```

PR #476, opened 2025-09-22
Make supplier and other array params mandatory

bryophyta:
## What does this change?

We have a number of search parameters in the UI that take arrays, which are currently optional. There's no clear meaning in this context for 'undefined', as opposed to an empty array, though, and we almost always expected the values to be arrays (empty or otherwise), so it makes sense to make sure it's always an array, so that the code is simpler (fewer checks to make sure it's an array) and easier to reason about.

## How to test

- [x] Run unit tests -- do they pass?
- [x] Make some searches using the relevant params via the URL and the UI -- are there any regressions?

### DIFF
--- a/newswires/client/src/urlState.test.ts
+++ b/newswires/client/src/urlState.test.ts
@@ -96,7 +96,6 @@ describe('urlToConfig', () => {
 			view: 'feed',
 			query: {
 				...defaultQuery,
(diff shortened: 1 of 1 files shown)


lindseydew (commented):
Nice, this is looking like a good shout 🙌 

Just had a question about the test set up

PR #484, opened 2025-09-24
Use ssm to manage db version scripts

lindseydew:
## What does this change?
This modifies the flyway script to include a step to persist the latest version applied in PROD to SSM. 

This also adds a shell script to check that version against the versions in the `migrations` folder, and fail the build on main if there has been a merged migration script that has not yet been applied to prod


.github/workflows/check-latest-migration.yaml
```diff
@@ -0,0 +1,24 @@
+name: check-latest-migration
+on:
+  pull_request:
```
lindseydew, line 3: I'm in two minds about filtering on the migrations path here. 

If we do filter and have this as a required check then PRs that don't change this folder will get a yellow status. I don't think it will actually block the merge, but it might feel a bit confusing to someone new to the code base. 

On the other hand, if we don't filter it will run for every pull request and most won't be making a migration change at all. I think the scope of the work means this isn't a big deal but open to suggestion on it.

PR #485, opened 2025-09-24
CDK changes for using ssm to manage database version

lindseydew:
## What does this change?
This adds an SSM paramater where we will keep track of the latest applied database versions, and a github role so that the CI script can access this parameter


cdk/lib/newswires.ts
```diff
@@ -469,5 +478,31 @@ export class Newswires extends GuStack {
 		newswiresApp.autoScalingGroup.connections.addSecurityGroup(
 			database.accessSecurityGroup,
 		);
+
+		if (this.stage === 'PROD') {
```
bryophyta, line 482: I wonder if it's changing this to `this.stage !== 'CODE'` or `['TEST', 'PROD'].includes(this.stage)` so that it shows up in the snapshots?

lindseydew: Yeah, I was thinking about the snapshots too. I've added a test check as well so we can see the cloudformation in the snapshot

PR #487, opened 2025-09-24
Order by

lindseydew:
## What does this change?
This adds sorting by ingestion time on the U.I. 

This is to handle a case where if we reprocess messages out of order that these will appear in the right place in the U.I

## Introducing DB migrations?


newswires/client/src/Feed.tsx
```diff
@@ -53,6 +53,12 @@ export const Feed = ({
 		return () => observer.disconnect();
 	}, [containerRef]);
 
+	const sortedResults = useMemo(() => {
```
bryophyta, line 56: nice! this works as expected when testing locally 👍

PR #489, opened 2025-09-30
Bump the threshold of dlq

lindseydew:
## What does this change?
The alarm we set is flickering on and off for the dead letter queue. 
dlq

## Introducing DB migrations?


lindseydew:
> I'm not sure if I understand the rationale for this yet. Do we know what's causing the number of messages in the DLQ to drop below 3 again, once it's gone up?

I was assuming some automatic redelivery. I mostly wanted to just make sure we were alerting only when there are issues to solve, and this alarm has been flickering red and green. 

But if this isn't the normal behaviour can do a deeper investigation and understand why we are getting these messages

bryophyta:
> But if this isn't the normal behaviour can do a deeper investigation and understand why we are getting these messages

I'm not sure tbh.. 🤔 So yeah it feels like it might be worth making sure we understand what's happening?

lindseydew:
Looking at the logs there are a few messages that go to the dead letter queue: 

logs

Sometimes we can't parse the json properly and sometimes there is email validation error. 

This is the messages visible over time. We have on average about 5. I think the variation could be due to the visibility window when the messages are being retried, or expiring retention

messages-visible

We currently have the dead letter queue set up to retain messages for 4 days. 
retention

If we don't want to be dropping any messages at all, we should investigate these and see if we need to tweak our ingestion lamba. 

If we expect a bit of variance, then I think upping the threshold a bit to ensure we catch it when things are broken would be a good course of action

PR #496, opened 2025-10-02
Precompute the ingestion lambda

lindseydew:
## What does this change?
This does a precomputation for category codes which have an especially long list. This is so that we can reduce load on the database for certain preset queries. 

## Introducing DB migrations?


shared/categories.ts
```diff
@@ -0,0 +1,1463 @@
...
+];
+export const AllBusiness = [
+	'apCat:f',
+    'paCat:FFF',
+    'paCat:GXX',
+    ...businessRelatedNewsCodes,
+    ...businessRelatedTopicCodes
+];
+export const Soccer = [
+    'N2:SOCC',
```
bryophyta, line 1392: there's a duplication here -- is it just duplicated or are there meant to be two variations?

lindseydew: Oh thanks, good spot

shared/precomputeCategories.ts
```diff
@@ -0,0 +1,26 @@
...
+    'sports-related-news-codes': (categoryCodes) => containsCode(categoryCodes, sportsRelatedNewsCodes),
+    'business-related-news-codes': (categoryCodes) => containsCode(categoryCodes, businessRelatedNewsCodes),
+    'business-related-topic-codes': (categoryCodes) => containsCode(categoryCodes, businessRelatedTopicCodes),
+    'other-topic-codes': (categoryCodes) => containsCode(categoryCodes, otherTopicCodes),
+}
+
+export const computePresetCategories = (categoryCodes: string[]) => {
+    return Object.entries(presetCodesMap).filter(([_, checkFn]) => {
+        return checkFn(categoryCodes);
+    }).map(([presetCategory]) => presetCategory);
```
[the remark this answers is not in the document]
lindseydew: Yeah that's nice :)

PR #500, opened 2025-10-07
Sports presets overhaul

lindseydew:
This represents sensible changes to Sport presets cherry-picked from [this PR](https://github.com/guardian/newswires/pull/465) (where multifield search sits) without multisearch needed to separate Soccer results for AFP (and useful in general).

## What does this change?

- Fixes sort order of Rugby presets in the dropdown
- Renames F1 to Motor Racing (and broadens its scope to all car racing), but without motorbikes, so maybe should be called Car Racing? Or include motorbikes?
- Adds Cycling Results presets to declutter cycling stories
- I went through all Sport presets, tightened them and separated results as best I could
- To improve it, we need two things I can’t see us having currently:
   - ability to specify a preset by searching in multiple fields at once (bad attempt [here](https://github.com/guardian/newswires/pull/465))
   - ability to exclude subsuppliers (`sourceFeed` field) (eg. some PA Rugby Results arrive as `pa:SSS`, but should be excluded by`sourceFeed:-"PA PA SPORT DATA"`)

## Potential further steps

- We should probably rename `All Data Formats` to `All Sport Tables`, but there is a broader question of making TYPES of stories more evident/accessible
- We should probably only show dataFormats in Soccer Tables preset, even before addressing the wider types question above (that’s why I haven’t added eg. AFP soccer tables to Soccer Tables)
- We should think if No Soccer should exclude (any) results (to only show real stories)

## How to test

Compare with current PROD.

This seems ready(ish) for human consumption. 

## Accessibility

### DIFF
--- a/newswires/app/conf/SearchPresets.scala
+++ b/newswires/app/conf/SearchPresets.scala
@@ -292,8 +292,11 @@ object SearchPresets {
 
   private val Cricket = List(
     SearchPreset
-      .fromSearchTerm(REUTERS, searchTerm = SearchTerm.Simple("-(OPTA)", SearchField.BodyText),
-        categoryCodes = CategoryCodes.Cricket.REUTERS),
+      .fromSearchTerm(
+        REUTERS,
+        searchTerm = SearchTerm.Simple("-(OPTA)", SearchField.BodyText),
+        categoryCodes = CategoryCodes.Cricket.REUTERS
+      ),
     SearchPreset(PA, CategoryCodes.Cricket.PA),
     SearchPreset
       .fromSearchTerm(AFP, searchTerm = SearchTerm.Simple("cricket", Slug), categoryCodes = CategoryCodes.Sport.AFP),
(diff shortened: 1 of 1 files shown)


newswires/app/conf/SearchPresets.scala
```diff
@@ -264,94 +275,140 @@ object SearchPresets {
...
+      CategoryCodes.RugbyUnion.REUTERS
+    ),
     SearchPreset.fromSearchTerm(
       PA,
       searchTerm = SearchTerm.Simple("RUGBYU", Slug),
-      categoryCodes = List("paCat:SRS", "paCat:SSS")
-    ),
-    SearchPreset.fromText(AFP, text = SimpleSearchQueries.RUGBY_UNION, CategoryCodes.Sport.AFP),
+      CategoryCodes.Sport.PA
+    ), // make excluding "sourceFeed": "PA PA SPORT DATA"
```
lindseydew, line 336: Let's remove this comment as well

bryophyta (approved):

PR #593, opened 2025-12-11
Upgrade app to Java 21

andrew-nowak:
## What does this change?

Upgrade newswires app to run on Java 21

## Introducing DB migrations?


bryophyta (approved):
Tested on CODE and it seems to work as expected to me 👍👍

PR #789, opened 2026-04-10
Fix #788: Remove unnecessary usage of Zod from type-definitions

### LINKED ISSUE #788: Many types are defined with Zod when they don't need to be
Our understanding is that [Zod](https://zod.dev/) is only needed for types that get transmitted from server to client - for instance, `WireDataFromAPI`.

However, it looks like Zod is getting used a lot more widely than that - for instance here, in the React context for the Search UI, which isn't transmitted client-server:  

https://github.com/guardian/newswires/blob/6e81a2b550321f6f4991d8f4841c93332259e08c/newswires/client/src/context/SearchContext.tsx#L46-L56

There's quite a lot of types that are Zod-defined, but it does make them harder to work with - you can't add any field unless Zod knows how to handle that type!

Discussed with the team here: https://chat.google.com/room/AAAAEGoouT4/_I_ExaXoCSc/_I_ExaXoCSc?cls=10

rtyley:
* https://github.com/guardian/newswires/issues/788

This PR uses typescript types instead of the `zod schema` definition in places where parsing is not needed

### DIFF
--- a/newswires/client/src/SearchSummary.tsx
+++ b/newswires/client/src/SearchSummary.tsx
@@ -143,8 +143,8 @@ const Summary = ({
 		(end === DEFAULT_DATE_RANGE.end || end === undefined);
 
 	const durationLabel = usePrettyDuration({
-		timeFrom: start ?? DEFAULT_DATE_RANGE.start,
-		timeTo: end ?? DEFAULT_DATE_RANGE.end,
+		timeFrom: start ?? (DEFAULT_DATE_RANGE.start as string),
+		timeTo: end ?? (DEFAULT_DATE_RANGE.end as string),
 		dateFormat: 'MMM D • HH:mm',
 	});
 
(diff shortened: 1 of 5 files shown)


newswires/client/src/SearchSummary.tsx
```diff
@@ -143,8 +143,8 @@ const Summary = ({
 		(end === DEFAULT_DATE_RANGE.end || end === undefined);
 
 	const durationLabel = usePrettyDuration({
-		timeFrom: (start ?? DEFAULT_DATE_RANGE.start) as string,
-		timeTo: (end ?? DEFAULT_DATE_RANGE.end) as string,
+		timeFrom: start ?? (DEFAULT_DATE_RANGE.start as string),
+		timeTo: end ?? (DEFAULT_DATE_RANGE.end as string),
```
bryophyta:
```suggestion
		timeFrom: start ?? DEFAULT_DATE_RANGE.start,
		timeTo: end ?? DEFAULT_DATE_RANGE.end,
```

bryophyta: I think we can drop the 'as string' entirely actually?

bryophyta (approved):
one v small comment but this lgtm! 🤩

PR #790, opened 2026-04-10
Upgrade PROD db to Large

bryophyta:
We've hit 100% CPU on the existing db instance, which has taken the site down. It looks like the CPU usage has been consistently high since we did the PA switch-over, presumably because the site is getting more usage. So it seems sensible to upsize it.

PR #791, opened 2026-04-10
Upgrade PROD db to XLarge

bryophyta:
We upgraded to Large in #790 before realising that Medium->Large doesn't increase CPU, which seemed to be the main bottleneck before. We'll see whether the extra memory from the previous upgrade helps, but if not we should probably go to XL.


rtyley:
RDS Instance types: https://aws.amazon.com/rds/instance-types/

image

andrew-nowak (approved):

PR #793, opened 2026-04-15
Add press release category to AFP stories and exclude from World

bryophyta:
## What does this change?

We receive fairly regular business press releases from AFP, which are currently showing up in 'World' preset, but which are not really of interest to users of that preset. 

* They tend to have the `ECO` category code, but lots of other more interesting stories have that code too.
* Their slugs aren't very predictable.
* However, their headlines begin `Press Release from Business Wire:`

I considered adding a single field text search exclusion in the preset definition, but it feels like a reasonable place at which to add our own category code instead, and filter the preset based on that? Currently, single field text searches are applied using Postgres' `english_unaccent` config, so it's harder to reason about exact text matching. And more generally I can see this being a category that we'd want to use more broadly for other suppliers, independently of the specific wording of the headlines in AFP.

## Introducing DB migrations?

### DIFF
--- a/newswires/app/conf/SearchPresets.scala
+++ b/newswires/app/conf/SearchPresets.scala
@@ -162,7 +162,11 @@ object SearchPresets {
   )
 
   private val AfpWorld = List(
-    SearchPreset(AFP, categoryCodesExcl = CategoryCodes.Sport.AFP)
+    SearchPreset(
+      AFP,
+      categoryCodesExcl = CategoryCodes.Sport.AFP :+ "guCat:PressRelease",
+      searchTerms = Some(SingleTerm(SearchTerm.SingleField("-\"Press Release from Business Wire:\"", Headline)))
+    )
   )
 
   private val MinorAgenciesWorld = List(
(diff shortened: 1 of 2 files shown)


newswires/app/conf/SearchPresets.scala
```diff
@@ -162,7 +162,11 @@ object SearchPresets {
   )
 
   private val AfpWorld = List(
-    SearchPreset(AFP, categoryCodesExcl = CategoryCodes.Sport.AFP)
+    SearchPreset(
+      AFP,
+      categoryCodesExcl = CategoryCodes.Sport.AFP :+ "guCat:PressRelease",
+      searchTerms = Some(SingleTerm(SearchTerm.SingleField("-\"Press Release from Business Wire:\"", Headline)))
```
andrew-nowak: why do we exclude both? If we're tagging on ingestion we shouldn't need the search term in this preset right?

andrew-nowak (approved):

newswires/app/conf/SearchPresets.scala
bryophyta: good shout, that was just left in from an earlier bit of testing! will remove now

PR #794, opened 2026-04-15
Allow sport and finance in US news if it also has US domestic codes

bryophyta:
## What does this change?

Remove the exclusion of finance (`apCat:f`) and sport (`apCat:s`) from the US news preset for AP. There aren't an overwhelming number of extra stories that have these codes in addition to the US domestic codes that we're including. And many of the extra stories are quite relevant to US editors, according to conversations we've had.

## How to test

- [x] Run locally against the CODE db, and compare with actual CODE. Are the extra stories broadly relevant?

## Introducing DB migrations?

### DIFF
--- a/newswires/app/conf/SearchPresets.scala
+++ b/newswires/app/conf/SearchPresets.scala
@@ -194,8 +194,7 @@ object SearchPresets {
   private val AllUs = List(
     SearchPreset(
       AP,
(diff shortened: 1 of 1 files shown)


lindseydew (approved):
LGTM 👍 

Just looks like some formatting is needed

PR #796, opened 2026-04-17
Pf/spike category code and

bryophyta:
## Introducing DB migrations?


bryophyta:
Closed in favour of alternative implementation #797

PR #797, opened 2026-04-20
All/Some combinator for category codes

bryophyta:
## What does this change?

Add a wrapper to the list of category codes for `categoryCodesIncl` which specifies whether the category codes need to all be present, or just some of them (the difference is [between `@>` and `&&`](https://www.postgresql.org/docs/current/functions-array.html#FUNCTIONS-ARRAY) in the ultimate Postgres query).

The implementation follows the pattern we've got for `SearchTerm`. All existing uses of `categoryCodesIncl` have been given `SOME`, which was implicitly the only option before. This PR should be a no-op refactor, but we will add some uses of `ALL` in follow-up PRs.

## How to test

- [x] Run locally, do some searches involving presets and category code inclusion/exclusion. Compare the same searches with CODE; they should be the same.

## Introducing DB migrations?

### DIFF
--- a/newswires/app/conf/CategoryCodesCondition.scala
+++ b/newswires/app/conf/CategoryCodesCondition.scala
@@ -1,0 +1,6 @@
+package conf
+
+case class CategoryCodesCondition(
+    categoryCodes: List[String],
+    combiner: Combination
+)
--- a/newswires/app/conf/CategoryCodesCondition.scala
+++ b/newswires/app/conf/CategoryCodesCondition.scala
@@ -1,6 +1,12 @@
 package conf
 
+sealed trait SetCombination
+
+case object SOME extends SetCombination
+case object ALL extends SetCombination
+case object NONE extends SetCombination
+
 case class CategoryCodesCondition(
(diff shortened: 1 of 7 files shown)


newswires/app/conf/CategoryCodesCondition.scala
```diff
@@ -0,0 +1,11 @@
+package conf
+
+sealed trait SetCombination
+
+case object SOME extends SetCombination
```
lindseydew, line 10: These names make sense now that I've read the docs for the sql operation they relate to: https://www.postgresql.org/docs/current/functions-array.html#FUNCTIONS-ARRAY

Am wondering if a comment or something next to the `SetCombinator` to explain the distinction may be clarifying?

Am also a bit wary of code comments, so happy for push back on this.

lindseydew (approved):
Looks great, nice work

newswires/app/conf/CategoryCodesCondition.scala
bryophyta: good call, I've done that now 👍

PR #798, opened 2026-04-20
Add typed timezone support

lindseydew:
## What does this change?

This introduces two new interfaces:
`InstantMoment` and `ZonedMoment` 

The `InstantMoment` is to be used when we parse a raw `ingestedAt` timestamp, to represent a moment in time, with the only accessible method `toZonedMoment` which takes in the timezone as a dependency. 

This ensures we can never read the raw date out directly on the U.I, but we can access it with a timezone when one is available in the context. 

`ZonedMoment` now handles:
- converting utcTime to a timezoned version
- format logic for all the variations where we display a date. 

## Introducing DB migrations?


newswires/client/src/utils/date/ZonedMoment.ts
```diff
@@ -0,0 +1,51 @@
...
+	#utcTime: Moment;
+	#timezone: TimezoneId;
+	constructor(utcTime: Moment, timezone: TimezoneId) {
+		this.#utcTime = utcTime;
+		this.#timezone = timezone;
+	}
+	private zoned() {
+		return this.applyTimezone(this.#utcTime);
+	}
+	private applyTimezone(m: Moment) {
```
bryophyta, line 21: It looks like we can use the `#` prefix for methods, too, if we want to enforce privacy at runtime? (not blocking as we're using TypeScript throughout anyway, but the stricter requirement feels sensible?)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements

newswires/client/src/utils/date/ZonedMoment.ts
```diff
@@ -0,0 +1,51 @@
...
+				return m.format('HH:mm');
+			case 'contextAware': {
+				const nowM = this.applyTimezone(context.nowUtc);
+				const sameDay = m.isSame(nowM, 'day');
+				return m.format(sameDay ? 'HH:mm:ss' : 'YYYY/MM/DD, HH:mm:ss');
+			}
+			default: {
+				const _exhaustive: never = context;
+				return _exhaustive;
+			}
```
bryophyta, line 48: It looks like this default case is intended to satisfy an exhaustiveness check, rather than something that we're expecting to actually hit. Is that right? 

If so, where does the need come from? It seems to pass type checking and linting without this default case

newswires/client/src/utils/date/ZonedMoment.ts
```diff
@@ -0,0 +1,51 @@
...
+		switch (context.type) {
+			case 'humanFull':
+				return m.format('MMM Do YYYY, HH:mm:ss');
+			case 'isoString':
+				return m.format('YYYY-MM-DDTHH:mm:ssZ');
+			case 'relative':
+				return m.fromNow();
+			case 'shortTime':
+				return m.format('HH:mm');
+			case 'contextAware': {
```
bryophyta, line 40: I couldn't guess from the name `contextAware`, so I had to come and look at the code in order to work out what `contextAware` would return. I can't think of a name that I think is more descriptive though, so not blocking at all!

bryophyta (approved):
This looks great to me, and makes it much easier to track the values and to know how to add new usages 👍 

From testing locally it all seems to behave as I'd expect, although here are a lot of potential cases so I'm not sure if I've caught them all.

I've added a couple of comments/suggestions but I don't think anything blocking.

newswires/client/src/utils/date/ZonedMoment.ts
lindseydew: It's to force a compile error if you don't have an exhaustive list listed in the switch statement. So that if a new type is added to `FormatContext` it has to be added to the switch statement

If you comment out one of the switch statements, then it would fall into the default block. 

And this would force a compile error because you can't set a type to a never.

newswires/client/src/utils/date/ZonedMoment.ts
lindseydew: yeah, agree, it's a bit opaque. It kind of wants to be like `timeOrDateTime` or something? or `timeWithOptionalDate`?

PR #801, opened 2026-04-21
Simplify panel navigation: panel switching requires changing presets

bryophyta:
This allows us to pre-emptively fix some react-hooks linting warnings, before updating our linting config. It also fixes a bug whereby if you:

* had Sport selected and the first panel open,
* then navigate to the sports sub-panel
* it will have deselected Sport (expected behaviour: Sport stays selected and just the panel changes)

Extracted from #777


newswires/client/src/hooks/usePrevious.ts
```diff
@@ -1,9 +1,13 @@
...
-	return ref.current;
+export function usePrevious<T>(value: T): T | null {
+	const [current, setCurrent] = useState(value);
+	const [previous, setPrevious] = useState<T | null>(null);
+
+	if (value !== current) {
+		setPrevious(current);
+		setCurrent(value);
+	}
+
```
lindseydew: Just curios about the change here?

To my (admittedly not incredibly deep understanding), the previous implementation is slightly preferable because it's robust between rerenders

newswires/client/src/SideNav/PresetsContextMenu.tsx
```diff
@@ -36,35 +36,31 @@ export const PresetsContextMenu = () => {
...
 
-	useEffect(() => {
-		const nextPanel = getPresetPanel(activePreset);
-		if (nextPanel !== activePanelId && previousPreset !== activePreset) {
-			startAnimation(nextPanel);
-		}
-	}, [activePreset, previousPreset, activePanelId, startAnimation]);
+	if (
+		nextPanel !== activePanelId &&
+		previousPreset !== null &&
```
lindseydew: I'm actually not sure if we need to track the previous preset at all now? Now that the animation should only ever happen when we switch between different lists, I think it should be safe to deprecate the `previousPreset` tracking and conditional from what I've observed quickly locally. 

Though, may have missed something more fundamental

lindseydew (commented):
Nice, looking great. 

Thanks for splitting into a separate PR, made the review much easier 👍 

Just had a thought about a potential extra simplification, but nice one on seeing a simpler path in general

newswires/client/src/hooks/usePrevious.ts
bryophyta: Sorry I'd put a comment on this [in the original PR](https://github.com/guardian/newswires/pull/777#discussion_r3029203694) but didn't copy it over to here.

To be honest, I didn't go super deep into this change -- I checked the implementation in [useHooks](https://github.com/uidotdev/usehooks), which I usually find to be a good resource, and it didn't face the same linting error, so I thought it would do the trick.

I can see that the implementation in that package changed from a ref-based one previously, but unfortunately the commit doesn't explain why. I'd be happy to spend a bit of time trying to understand the logic together at some point if you're interested? https://github.com/uidotdev/usehooks/commit/c065ab1f9b7cb7f6dcd6f48f2054b95d98cf55b6

newswires/client/src/SideNav/PresetsContextMenu.tsx
bryophyta: hm.. good question!

i tried removing it at some point when i was working on this and found that it caused an issue, but it doesn't seem to happen anymore from what i can tell, so maybe some other change i made in the meantime was responsible for the issue earlier..?

i've just pushed a commit which removes this check (and the hook, as this was the only usage) -- what do you think?

lindseydew: Yeah, I pulled this down and tested locally and looks like it's all working ok 👍

newswires/client/src/hooks/usePrevious.ts
lindseydew: Looks like we are able to remove it anyway 👍

newswires/client/src/SideNav/TopLevelListPreset.tsx
```diff
@@ -1,6 +1,5 @@
 import { EuiListGroup } from '@elastic/eui';
 import { useSearch } from '../context/SearchContext';
-import { shouldTogglePreset } from '../presetHelpers';
```
lindseydew, line 3: Nice, I think we can delete this method and the tests altogether now

newswires/client/src/SideNav/PresetsContextMenu.tsx
```diff
@@ -34,37 +33,27 @@ export const PresetsContextMenu = () => {
...
 	const activePreset = config.query.preset;
-	const previousPreset = usePrevious(activePreset);
 
-	const [activePanelId, setActivePanelId] = useState<PresetGroupName>(
+	const [activePanelId, setActivePanelId] = useState<PresetGroupName>(() =>
 		getPresetPanel(activePreset),
 	);
 	const getPreviousPanelId = (panel: PresetGroupName): PresetGroupName =>
 		panel === 'presets' ? 'sportPresets' : 'presets';
 	const previousPanelId = getPreviousPanelId(activePanelId);
```
lindseydew: Small refactoring opportunity:

`const previousPanelId = activePanel === 'presets' ? 'sportPresets' : 'presets'`

lindseydew (approved):
Nice, looks greats 👍 

Just a couple of small non-blocking refactoring suggestions

newswires/client/src/SideNav/TopLevelListPreset.tsx
bryophyta: good call, done now!

PR #830, opened 2026-04-23
Remove semver-specific cooldown option from github-actions

andrew-nowak:
github-actions ecosystem does not allow setting options which requires semver understanding https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#configuration-of-cooldown


bryophyta (approved):
🙌

PR #833, opened 2026-04-23
Remove local stack

lindseydew:
## What does this change?
This removes the local stack configuration in favour of using `in memory` services to emulate connecting to `s3` and `sqs` when running locally. 

This also rewrites the config module to lazy load configuration objects and allow variables to be `optional` in dev mode and required in `code` or `prod` mode. 

## How to test

Tested in CODE that 

- [x] Ingestion lambda still runs
- [x] Fingerpost Queuing lambda still runs
- [x] Sending Email still works
- [x] Reuters poller lambda
- [x] AP Poller Lambda

## Images

logoutput
logoutput2

## Introducing DB migrations?


fingerpost-queueing-lambda/localRun.ts
```diff
@@ -1,129 +1,41 @@
...
-//   }
+export function localRunFingerpostQueueingLambda() {
+	const dummyEvent = createDummySQSEvent()
+	main(dummyEvent).then(console.log).catch(console.error);
+}
```
bryophyta: it doesn't look like this function is being used anywhere?

shared/src/config.ts
```diff
@@ -1,30 +1,132 @@
...
-	!process.env.LAMBDA_TASK_ROOT && !process.env.AWS_EXECUTION_ENV;
-
-export function getFromEnv(key: string): string {
-	const value = process.env[key];
-	if (!value) {
-		throw new Error(`Missing required environment variable ${key}`);
-	}
-	return value;
+type AppMode = 'dev' | 'code' | 'prod';
+const allowedAppModes: readonly AppMode[] = ['dev', 'code', 'prod'];
```
bryophyta:
```suggestion
const allowedAppModes = ['dev', 'code', 'prod'] as const;
type AppMode = (typeof allowedAppModes)[number];
```

suggestion for removing the need to duplicate the list

shared/src/config.ts
```diff
@@ -1,30 +1,132 @@
...
+}
+
+export const appConfig = buildAwsConfig();
+export const databaseConfig = () => buildDatabaseConfig();
+export const ingestionQueueUrl = () =>
+	envForStage('INGESTION_LAMBDA_QUEUE_URL') ?? 'dummy-ingestion-queue';
+export const feedsBucket = () =>
+	envForStage('FEEDS_BUCKET_NAME') ?? 'dummy-feeds-bucket';
+export const emailBucket = () =>
+	envForStage('EMAIL_BUCKET_NAME') ?? 'dummy-email-bucket';
```
bryophyta: I think the logic here makes sense, but I found the name `envForStage` didn't really help me understand what was happening. I'd almost be tempted to suggest something like:

```
APP_MODE === 'dev' ? env('key').orElse('default_val') : env('key').required()
```

but this isn't a blocker I don't think; happy to defer to what feels right to you

shared/src/s3.ts
```diff
@@ -5,94 +5,167 @@ import {
...
-	bucketName,
-	key,
-}: {
-	bucketName: string;
-	key: string;
-}): Promise<OperationResult<{ body: string; lastModified?: Date }>> {
-	logger.log({
-		message: `Getting object from S3 bucket "${bucketName}" with key "${key}"`,
+interface FileService {
+	putToS3({
```
bryophyta: this is a tiny point, but: I'd be tempted to rename these methods to just `putObject` and `getObject`, or something like that, given that they can be backed by something other than S3 now

shared/src/s3.ts
```diff
@@ -5,94 +5,167 @@ import {
...
-			reason: getErrorMessage(caught),
+		key,
+		body,
+	}: {
+		bucketName: string;
+		key: string;
+		body: string | Buffer;
+	}): Promise<OperationResult<{ response: PutObjectCommandOutput }>> {
+		const now = new Date();
+		const bodyStr = typeof body === 'string' ? body : body.toString();
```
bryophyta, line 127: do we need to pass a `Buffer` currently? If not, I'd be tempted to narrow `body` to just take `string`, unless we know how the S3 sdk handles buffer serialisation?

bryophyta (approved):
This looks great to me!

I've left a few comments about implementation detail, and we discussed the LocalRun async stuff yesterday.

Overall though I think this makes for a much clearer way of handling config overall throughout the app!

shared/src/s3.ts
lindseydew: Yeah it probably is a bit overkill. But it aligns with the current implementation: https://github.com/guardian/newswires/blob/main/shared/src/s3.ts#L73

so I think it would be better to leave it the same and then maybe refactor as a seperate step

PR #834, opened 2026-04-24
Add 'data:finance' to the list of PA Business codes

bryophyta:
We've got business stories arriving which don't have any of the previously recognised PA business category codes, but which do have `data:finance`, so we should add them into the Business presets (and thereby take them out of Sport, which is where they were showing up by default 😮)


paperboyo (approved):
Thx!

PR #836, opened 2026-04-24
Wire up UkElection preset, shown in nav only to dev users

bryophyta:
## What does this change?

We started getting a lot of election test stories today, which were ending up in Sport, but filtered out in #835. As it stands, this leaves them with no alternative home. We might want to add them to the UK preset (for UK elections) or create a separate preset, but for the time being it'll probably be useful for us to be able to see them while we're developing.

Depends on the list of dev users (developers and other team members) that we use to enable us to see which telemetry events are coming from our own testing. This feels like a reasonable group/mechanism to use for this purpose, too.

## Introducing DB migrations?


paperboyo (approved):
Thx! Assuming my approval is of use.

PR #841, opened 2026-04-27
Test alternative to 'not exists' queries

bryophyta:
## New query logic

Change `exclusionCondition` so that now takes a `queryVariant` parameter.

* Given `not_exists` it will produce the same query as before, using `NOT EXISTS (SELECT FROM ...)`.
* Given `plain_not` (or if unspecified) it will just wrap the provided clause in `NOT (...)`

Testing in CODE shows that the new variant speeds up the most expensive queries dramatically (`World + UK, 1 Week` goes down from ~17s to ~1s).

The reason for this seems to be that the `NOT EXISTS` queries need to do joins across a lot of rows before it selects the top 30. The 'plain not' variant tends to inline the checks as it loops through candidate rows, knowing that it's likely to be able to exit early once it's found 30 matching items.

## 'a/b test' implementation

We've stress tested this a bit in CODE, and found that generally performance with the new variant seems to be about the same as, or in some cases significantly better than, the old variant.

However, we theorised that complex queries with sparse results might be slower, and @paperboyo found at least one case which is significantly slower in the new vs. the old variant: text search for an uncommon term within the `No Sport` preset. (Text search within `has data formatting` was catastrophically slow, but seems to have been resolved by adding an index.)

It's difficult to map out all the implications with confidence in advance, so I'm proposing that we do an 'a/b test' where the server will pick a query variant at random, passing it through to the client so that the variant can be included in our performance telemetry. This should allow us to compare the two strategies based on a larger amount of real user data.

## Count query changes

The speed-up for complex and long-range queries is largely due to the db not needing to process all matching entries when it only needs 30. This doesn't apply to a count query, however! Slowing the user experience down significantly just for the sake of getting an accurate count doesn't seem right.

I've experimented with some ways of improving the count query performance, but haven't had any success yet. On discussion with the team, we decided that it would be best to get this change tested and out, with a simple cap on the count query. Then, in the future, we can look into either improving the performance of the count, or loading it asynchronously on the client after the results are fetched.

## Todo

- [x] Test how it performs for other presets
- [x] Check interaction with custom filters
- [x] See whether the `NOT EXISTS` -> `NOT` conversion makes sense in every case, or just some.
- [x] Reconcile the UI with the fact that we aren't expecting full counts anymore (or find a way to optimise the original count query)
- [ ] Check that the performance against CODE isn't too influenced by the fact that I bumped the statistics for the category_codes column from ~100 to ~600.

### DIFF
--- a/newswires/app/db/FingerpostWireEntry.scala
+++ b/newswires/app/db/FingerpostWireEntry.scala
@@ -30,6 +30,7 @@ import play.api.Logging
 import scalikejdbc._
 import io.circe.parser._
 
+import java.io.{File, FileWriter}
 import java.time.Instant
 
 case class WireMaybeToolLinkAndCollection(
@@ -58,6 +59,8 @@ case class FingerpostWireEntry(
 object FingerpostWireEntry
     extends SQLSyntaxSupport[FingerpostWireEntry]
     with Logging {
+
+  val countQueryCap = 100L
 
   implicit val jsonEncoder: Encoder[FingerpostWireEntry] =
     deriveEncoder[FingerpostWireEntry].mapJson(_.dropNullValues)
@@ -204,23 +207,33 @@ object FingerpostWireEntry
       .flatten
   }
 
+  private def formatQueryParam(param: Any): String = param match {
+    case null | None   => "NULL"
+    case Some(v)       => formatQueryParam(v)
+    case list: List[_] =>
+      "ARRAY[" + list
+        .map(e => "'" + e.toString.replace("'", "''") + "'")
+        .mkString(", ") + "]::text[]"
+    case n: Int  => n.toString
+    case n: Long => n.toString
+    case other   => "'" + other.toString.replace("'", "''") + "'"
+  }
+
+  private def _toExecutableQuery(
+      statement: String,
+      parameters: Seq[Any]
+  ): String = {
+    val sentinel = "\u0000"
+    val segments = statement.replace("??", sentinel).split("\\?", -1)
+    val formattedParams = parameters.map(formatQueryParam) :+ ""
+    segments
+      .zip(formattedParams)
+      .map { case (seg, param) => seg + param }
+      .mkString
+      .replace(sentinel, "?") // restore PostgreSQL ?| / ?& operators
+  }
+
   object Filters {
-    private def exclusionCondition(
-        alias: QuerySQLSyntaxProvider[SQLSyntaxSupport[
-          FingerpostWireEntry
-        ], FingerpostWireEntry]
-    )(innerClause: SQLSyntax) = {
-      // unpleasant, but the sort of trick you need to pull
-      // because "NOT IN (...)" doesn't hit an index.
-      // https://stackoverflow.com/a/19364694
-
-      sqls"""|NOT EXISTS (
-             |  SELECT FROM ${FingerpostWireEntry as alias}
-             |  WHERE ${syn.id} = ${alias.id}
-             |    AND $innerClause
-             |)""".stripMargin
-    }
-
     private def supplierCondition(
         alias: QuerySQLSyntaxProvider[SQLSyntaxSupport[
           FingerpostWireEntry
@@ -276,8 +289,7 @@ object FingerpostWireEntry
 
     lazy val supplierExclSQL =
       (suppliersExcl: List[String]) => {
-        val se = syntax("suppliersExcl")
-        exclusionCondition(se)(supplierCondition(se, suppliersExcl))
+        sqls"NOT (${supplierCondition(syn, suppliersExcl)})"
       }
 
     lazy val guSourceFeedSQL: List[String] => SQLSyntax =
@@ -289,12 +301,9 @@ object FingerpostWireEntry
 
     lazy val guSourceFeedExclSQL =
       (guSourceFeedsExcl: List[String]) => {
-        val gsfe = syntax("guSourceFeedExcl")
-        exclusionCondition(gsfe)(
-          sqls.in(
-            sqls"upper(${gsfe.guSourceFeed})",
-            guSourceFeedsExcl.map(feed => sqls"upper($feed)")
-          )
+        sqls.notIn(
+          sqls"upper(${syn.guSourceFeed})",
+          guSourceFeedsExcl.map(feed => sqls"upper($feed)")
         )
       }
 
@@ -349,8 +358,7 @@ object FingerpostWireEntry
 
     lazy val keywordsExclSQL =
       (keywords: List[String]) => {
-        val ke = syntax("keywordsExcl")
-        exclusionCondition(ke)(keywordCondition(ke, keywords))
+        sqls"NOT (${keywordCondition(syn, keywords)})"
       }
 
     lazy val categoryCodeInclSQL =
@@ -364,10 +372,7 @@ object FingerpostWireEntry
 
     lazy val categoryCodeExclSQL =
       (categoryCodesExcl: List[String]) => {
-        val cce = syntax("categoryCodesExcl")
-        exclusionCondition(cce)(
-          categoryCodeSomeConditions(cce, categoryCodesExcl)
-        )
+        sqls"NOT ${categoryCodeSomeConditions(syn, categoryCodesExcl)}"
       }
 
     lazy val dataFormattingSQL =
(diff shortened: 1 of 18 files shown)


newswires/test/db/FingerpostWireEntrySpec.scala
```diff
@@ -635,9 +635,7 @@ class FingerpostWireEntrySpec extends AnyFlatSpec with Matchers with models {
 
   it should "create the correct sql snippet for suppliersExcl" in {
```
lindseydew, line 619: I know this is intended for a short term test, but I think it would be worth including in at least one of the tests the sql that gets generated when the other flag it set

newswires/client/src/SearchSummary.tsx
```diff
@@ -258,24 +258,35 @@ const Summary = ({
 	);
 };
 
+function decideSearchSummaryLabel(
+	queryData: WiresQueryData | undefined,
+): string {
+	if (queryData === undefined || queryData.totalCount === 0) {
+		return 'No results found';
+	}
+	if (queryData.totalCount > queryData.countQueryCap) {
```
lindseydew: Should this equality be the other way around? The totalCount would be capped at the countQueryCap?

newswires/client/src/SearchSummary.tsx
```diff
@@ -258,24 +258,35 @@ const Summary = ({
...
+		return `Showing ${queryData.countQueryCap}+ results`;
+	}
+	return `Showing ${Intl.NumberFormat('en-GB').format(queryData.totalCount)} result${queryData.totalCount > 1 ? 's' : ''}`;
+}
+
 export const SearchSummary = ({
 	setSideNavIsOpen,
+	queryData,
```
lindseydew: I'm a bit confused as to why the queryData is being passed in as a property now rather than reading from the state?

newswires/client/src/context/SearchReducer.test.ts
```diff
@@ -25,6 +25,7 @@ describe('SearchReducer', () => {
 		queryData: {
 			results: [{ ...sampleWireData }],
 			totalCount: 1,
+			countQueryCap: 100,
```
lindseydew: Small point, but there seems to be a lot of recurrences of: 

```
      results: [{ ...sampleWireData }],
			totalCount: 1,
			countQueryCap: 100,
```
in this file and I wonder whether this could be abstracted into it's own object?

db/migrations/V33__Create_hasDataFormatting_index.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX CONCURRENTLY IF NOT EXISTS fingerpost_wire_entry_ingested_at_dataformat_idx
```
lindseydew, line 1: Small point, but I would suggest creating a PR for the migration separately and deploying this first. 

Just in case anything in this PR needs to be rolled back, and then you don't need to worry about migrations syncing as well.

newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -558,7 +562,8 @@ object FingerpostWireEntry
       queryCursor: QueryCursor,
       queryOrdering: TimeStampColumn,
       searchPresets: List[FilterParams] = Nil,
-      negatedSearchPresets: List[FilterParams] = Nil
+      negatedSearchPresets: List[FilterParams] = Nil,
+      queryVariant: QueryVariant = PlainNot
```
lindseydew, line 566: Up to you on this one, but it might be preferable not to use a default variable on methods that now require a `queryVariant` just to force a compiler error if the calling method hasn't actually passed the param in to make sure what is set at the controller is passed all the way down the chain. 

(Might impact on other things calling it like tests, so no worries if this doesn't feel worth the effort)

newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -647,7 +652,9 @@ object FingerpostWireEntry
   }
 
   def query(
-      queryParams: QueryParams
+      queryParams: QueryParams,
```
lindseydew, line 655: I would be tempted to include a `back door` count that does show the total count for a selection of filters that could be queried on the API by a developer. 

Could either make the field overridable at the query API level, or fork the query and make a new endpoint.

newswires/client/src/SearchSummary.tsx
lindseydew: Might be good to write a unit test on this method now that the display logic is slightly more complicated

newswires/app/db/FingerpostWireEntry.scala
```diff
@@ -683,15 +690,18 @@ object FingerpostWireEntry
...
+            |   SELECT 1
+            |   FROM ${FingerpostWireEntry as FingerpostWireEntry.syn}
+            |   LEFT JOIN ${ToolLink as ToolLink.syn}
+            |    ON ${syn.id} = ${ToolLink.syn.wireId}
+            |   LEFT JOIN ${WireEntryForCollection as WireEntryForCollection.syn}
+            |     ON ${WireEntryForCollection.syn.wireEntryId} = ${syn.id}
+            |   LEFT JOIN ${Collection as Collection.syn}
+            |     ON ${Collection.syn.id} = ${WireEntryForCollection.syn.collectionId}
+            |   WHERE $whereClause
+            |   LIMIT ${countQueryCap + 1}
```
lindseydew, line 703: Also could be nice to add a unit test on the count query to verify it includes the limit clause, but not a blocker

newswires/client/src/context/SearchContext.tsx
```diff
@@ -599,3 +591,35 @@ export const useSearch = () => {
...
+				JSON.stringify(value),
+			]),
+		),
+		duration: performance.now() - startedAt,
+		resultsCount: data.results.length,
+		resultsIds: data.results.map((wire) => wire.id).join(','),
+		totalCount: data.totalCount,
+		isRefresh,
+		...Object.fromEntries(
+			Object.entries(data.queryVariant ?? {}).map(([key, value]) => [
```
lindseydew, line 619: Nice, good tracking ✨

lindseydew (commented):
Looking great, great initiative 💯

db/migrations/V33__Create_hasDataFormatting_index.sql
bryophyta: good call! done now in #867

newswires/client/src/SearchSummary.tsx
bryophyta: The way I've done it is:

`countQueryCap = n; LIMIT in query = n + 1; therefore, if total > countQueryCap we say "n+ results"`

This is to allow us to say `"n results"` rather than `"n+ results"` in the case where there are exactly `n`.

A unit test seems like a good idea, and I'll add an inline comment to this line too because I agree it reads a bit counterintuitively

newswires/client/src/SearchSummary.tsx
bryophyta: no good reason I don't think! addressed now

newswires/client/src/context/SearchReducer.test.ts
bryophyta: done 👍

newswires/app/db/FingerpostWireEntry.scala
bryophyta: I thought about this too, but yes I decided the diff on all the tests would get very large if so. Given that the existence of this param is temporary it didn't feel like the right balance to me

newswires/test/db/FingerpostWireEntrySpec.scala
bryophyta: done now in ce2c5942 👍

newswires/app/db/FingerpostWireEntry.scala
bryophyta: Given that the count query is constructed inline in the `query` method, rather than being constructed via a dedicated function, it's not exposed in a way that lends itself to testing at the moment. My feeling is that if we introduced more complex logic into the construction of the count query then it might be worth extracting and adding tests at a later point, but it doesn't feel urgent at the moment to me, given that it's being constructed in a relatively declarative way for now?

newswires/app/db/FingerpostWireEntry.scala
bryophyta: good call; I've just added a `countQueryCap` override to the API endpoint for now in da7a0f1b

I could see us wanting to expose this option on the client at some point too, but feels like that would be better for a follow-up PR after we've had a broader conversation about the direction for this?

newswires/app/db/FingerpostWireEntry.scala
lindseydew: Yep, I think that's fine. We can always come back to this further down the line once the idea is a bit more solid. 

I think if we do end up with capping the count query it would be helpful at some point to have a unit test documenting this, so that if the limit ever changed we would have to explicitly update a test for this. 

But since we're still proving out the idea, we can come back to it later

lindseydew (approved):
Looks great 👍

PR #847, opened 2026-05-01
Show elections feed to all users

bryophyta:
Shows the 'Election UK' preset to all users. We've got local elections coming up next week, so we're planning to show the preset for a few days either side and then hide it again afterwards.


blishen (approved):
LGTM

PR #849, opened 2026-05-05
send followup telemetry event every 60s that user remains on search with auto-refresh on

andrew-nowak:
## What does this change?

Adds a periodic (~1 min) telemetry event while user remains on the results page/ticker with auto-refresh turned on. Should give a rough idea of "human-minutes spent on presets" in telemetry results.

## How to test

tested on CODE. Telemetry continues to fire when a preset is selected, and then roughly every minute afterwards.

## Introducing DB migrations?


lindseydew (approved):
LGTM 👍

PR #857, opened 2026-05-08
handy bash script for quickly logging in to remote dbs

andrew-nowak:
## What does this change?

From time to time we want to connect to the real database in psql, but remembering the right incantation is a bit of a faff. I've said for a while that I'd come back and script it, here it finally is

## Introducing DB migrations?


bryophyta (approved):
nice!! 🤩

PR #858, opened 2026-05-08
Add AP feature codes to US presets

hk-15:
## What does this change?
Adds AP codes for 'lifestyle', 'entertainment' and 'weather' to the US preset, which currently focuses on news stories.

Although news desks may not be interested in these stories, there are comparatively few of them. Without this change, anyone who is interested in these stories can only find them via a search or in the 'all' preset which contains 1000s of stories.

## How to test

Tested on CODE. Compared number of stories appearing for the feature codes with US preset selected and saw more on CODE than on PROD.

## Introducing DB migrations?

### DIFF
--- a/newswires/app/conf/SearchPresets.scala
+++ b/newswires/app/conf/SearchPresets.scala
@@ -209,7 +209,7 @@ object SearchPresets {
   private val AllUs = List(
     SearchPreset(
       AP,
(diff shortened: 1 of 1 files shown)


andrew-nowak (approved):
LGTM!

bryophyta (approved):
🥳

PR #859, opened 2026-05-08
Don't return 501 in response to requests with an unsupported method

bryophyta:
## What does this change?

There are a bunch of methods which we don't support, and the app is currently returning 501s for any requests with these, as per the HTTP spec. We have alarms based on the number of 5XXs, which normally indicate something has gone wrong. However, we keep getting spammed with unsupported methods which is making our alarms noisy for something that's not an actionable issue. 

See https://github.com/guardian/grid/pull/4528 for more info on considerations and implementation.

## How to test

Deploy to CODE and `curl` the load balancer *directly* with an unrecognised method (e.g. `-X PROPFIND` (requests to the dev-gutools.co.uk URL will be handled by Cloudfront before they hit the new code path). You should get a 400 instead of a 501 in response.

## Introducing DB migrations?

### DIFF
--- a/newswires/app/lib/UnknownMethodParsingErrorHandler.scala
+++ b/newswires/app/lib/UnknownMethodParsingErrorHandler.scala
@@ -1,0 +1,43 @@
+package lib
+
+import org.apache.pekko.event.LoggingAdapter
+import org.apache.pekko.http.scaladsl.model.{
+  ErrorInfo,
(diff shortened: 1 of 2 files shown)


andrew-nowak (approved):

lindseydew (approved):
LGTM 👍 Nice initiative

PR #861, opened 2026-05-11
Remove temporary filtering of legacy PA source feeds

andrew-nowak:
## What does this change?

Remove the PA API preview logic, now that the switchover is complete. We shouldn't be getting anything in from these source feeds any longer, but if anything should pop up then we can allow it to be displayed, as its presumbly something we haven't considered (like election night results :( )

## Introducing DB migrations?


lindseydew (approved):
LGTM 👍

PR #873, opened 2026-05-15
Enhance telemetry with server information

bryophyta:
## What does this change?

Add some fields to our client side telemetry:

- Pass a request id as a header to the server when making `fetch` requests, which as of #871 should be picked up and used by the server's own logging.
- Include that request id in our `NEWSWIRES_FETCHED_RESULTS` telemetry logging, allowing us to correlate the telemetry from the client with logging from the server.
- Pass a `Server-Timing` header back from the client which includes the server-side request duration, using [the standardised format](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Server-Timing) that developer tools know how to interpret.
- Extract the server timing from the response, and include it in our telemetry, so that now we have `duration` (total round trip from making the request to processing the response, in the client app) and `serverTiming` (the time spent processing the request on the server).

## How to test

- [x] Added some unit tests to the `extractServerTiming` function, to make sure that it handles edge cases without throwing.
- [x] Deployed to CODE and verified that the new logging shows up (see screenshot below).
- [x] I've also tried making requests with the network being throttled on the client, so that there's a much longer total duration but a normal server time. We can see that there's a long `duration` and a short `serverTiming` in the logs, which suggests that it's tracking what it's supposed to be.

## Images

image

image

## Introducing DB migrations?

### DIFF
--- a/newswires/app/lib/RequestLoggingFilter.scala
+++ b/newswires/app/lib/RequestLoggingFilter.scala
@@ -39,17 +39,17 @@ class RequestLoggingFilter(implicit
 
     val resultFuture = next(withID)
 
-    resultFuture onComplete {
+    resultFuture.transform {
       case Success(response) =>
         val duration = System.currentTimeMillis() - start
         log(withID, Right(response), duration)
+        Success(response.withHeaders("Server-Timing" -> s"total;dur=$duration"))
 
       case Failure(err) =>
         val duration = System.currentTimeMillis() - start
         log(withID, Left(err), duration)
+        Failure(err)
     }
-
-    resultFuture
   }
 
   private def log(
--- a/newswires/client/src/context/SearchContext.tsx
+++ b/newswires/client/src/context/SearchContext.tsx
@@ -22,7 +22,7 @@ import {
 import { recognisedSuppliers } from '../suppliers.ts';
 import { configToUrl, defaultConfig, urlToConfig } from '../urlState.ts';
 import { takeWhile } from '../utils/takeWhile.ts';
-import { fetchResults } from './fetchResults.ts';
+import { fetchResults } from './fetchResults/fetchResults.ts';
 import {
 	loadOrSetInLocalStorage,
 	saveToLocalStorage,
--- a/newswires/client/src/context/SearchContext.tsx
+++ b/newswires/client/src/context/SearchContext.tsx
@@ -276,7 +276,7 @@ export function SearchContextProvider({ children }: PropsWithChildren) {
 			const startedAt = performance.now();
 			refreshesSinceLastTelemetrySend.current = 0;
 			fetchResults({ query: currentConfig.query, view: currentConfig.view })
-				.then((data) => {
+				.then(({ data, requestId }) => {
 					sendTelemetryEvent(
 						'NEWSWIRES_FETCHED_RESULTS',
 						createFetchedResultsTelemetryData({
@@ -284,6 +284,7 @@ export function SearchContextProvider({ children }: PropsWithChildren) {
 							data,
 							startedAt,
 							isRefresh: false,
+							requestId,
 						}),
 					);
 					setHasBeenVisibleItemIds([]);
@@ -306,7 +307,7 @@ export function SearchContextProvider({ children }: PropsWithChildren) {
 						abortController,
 						view: currentConfig.view,
 					})
-						.then((data) => {
+						.then(({ data, requestId }) => {
 							refreshesSinceLastTelemetrySend.current += 1;
 							// only send refresh telemetry events every 60s to avoid flooding the telemetry index
 							if (
@@ -321,6 +322,7 @@ export function SearchContextProvider({ children }: PropsWithChildren) {
 										data,
 										startedAt,
 										isRefresh: true,
+										requestId,
 									}),
 								);
 							}
@@ -493,7 +495,7 @@ export function SearchContextProvider({ children }: PropsWithChildren) {
 			beforeTimeStamp,
 			view: currentConfig.view,
 		})
-			.then((data) => {
+			.then(({ data }) => {
 				dispatch({ type: 'APPEND_RESULTS', data });
 
 				if (selectNextItem && data.results.length > 0) {
@@ -597,11 +599,13 @@ function createFetchedResultsTelemetryData({
 	data,
 	startedAt,
 	isRefresh,
+	requestId,
 }: {
 	query: Query;
 	data: WiresQueryData;
 	startedAt: number;
 	isRefresh: boolean;
+	requestId: string;
 }) {
 	return {
 		...Object.fromEntries(
@@ -615,6 +619,7 @@ function createFetchedResultsTelemetryData({
 		resultsIds: data.results.map((wire) => wire.id).join(','),
 		totalCount: data.totalCount,
 		isRefresh,
+		requestId,
 		...Object.fromEntries(
 			Object.entries(data.queryVariant ?? {}).map(([key, value]) => [
 				`query-variant_${key}`,
--- a/newswires/client/src/context/SearchContext.tsx
+++ b/newswires/client/src/context/SearchContext.tsx
@@ -276,7 +276,7 @@ export function SearchContextProvider({ children }: PropsWithChildren) {
 			const startedAt = performance.now();
 			refreshesSinceLastTelemetrySend.current = 0;
 			fetchResults({ query: currentConfig.query, view: currentConfig.view })
-				.then(({ data, requestId }) => {
+				.then(({ data, requestId, serverTiming }) => {
 					sendTelemetryEvent(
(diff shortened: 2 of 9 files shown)


newswires/client/src/context/fetchResults.ts
```diff
@@ -10,14 +12,21 @@ export const fetchResults = async ({
...
 	abortController?: AbortController;
-}): Promise<WiresQueryData> => {
+	requestId?: string;
+}): Promise<{
+	data: WiresQueryData;
+	requestId: string;
+	serverTiming?: number;
+}> => {
 	const endpoint = view.includes('dotcopy') ? '/api/dotcopy' : '/api/search';
+	const requestIdToUse = requestId ?? uuidv4();
```
lindseydew: I found it a bit confusing that that this request id is passed as a param with a dynamic default if not. 

Is the idea that a new uuid is generated when it's called in application logic, and passed in explicitly for testing? 

Might be better to generate the uuidv4() on invocation and mock for a test? 

But let me know if I am missing something

newswires/app/lib/RequestLoggingFilter.scala
```diff
@@ -39,17 +39,17 @@ class RequestLoggingFilter(implicit
 
     val resultFuture = next(withID)
 
-    resultFuture onComplete {
+    resultFuture.transform {
       case Success(response) =>
         val duration = System.currentTimeMillis() - start
         log(withID, Right(response), duration)
+        Success(response.withHeaders("Server-Timing" -> s"total;dur=$duration"))
```
lindseydew, line 46: Curious as to why the server is just sending the total time when parsing unit tests include a database time as well? Is this in preparation for a possible extension?

newswires/client/src/context/fetchResults.ts
```diff
@@ -46,8 +56,40 @@ export const fetchResults = async ({
...
+		requestId: requestIdToUse,
+		serverTiming,
 	};
 };
+
+function extractServerTiming(headers: Headers | undefined): number | undefined {
+	const serverTimingHeader = headers?.get('Server-Timing');
+	const serverTimingMatch = serverTimingHeader?.match(
+		/total;dur=(\d+(\.\d+)?)/,
+	);
```
lindseydew: Think this method may be a bit more extensible if it parsed the Headers into a key value pair instead. For example:

```

function parseServerTiming(header) {
	if (!header) return [];

	return header.split(',').map(entry => {
		const [name, ...params] = entry.trim().split(';');

		const result = { name };

		for (const param of params) {
			const [key, value] = param.split('=');

			result[key] = value?.replace(/^"|"$/g, '');
		}

		return result;
	});
}
```

If this intention is to include the db times at some point. 

If not though, might be simpler for the server to just return the time and simplify the header structure?

newswires/client/src/context/fetchResults.ts
```diff
@@ -46,8 +56,40 @@ export const fetchResults = async ({
...
+		return serverTiming;
+	} catch (e) {
+		console.error(
+			`Error extracting Server-Timing from response: ${getErrorMessage(e)}`,
+		);
+		return undefined;
+	}
+}
+
+export const forTesting = {
```
lindseydew: I would just export the method directly so that we don't have to worry about a testing branch in the application code

newswires/client/src/context/SearchContext.tsx
```diff
@@ -611,10 +619,12 @@ function createFetchedResultsTelemetryData({
 			]),
 		),
 		duration: performance.now() - startedAt,
+		...(serverTiming !== undefined ? { serverTiming } : {}),
```
lindseydew, line 622: Slightly simpler: 
```
...(serverTiming !== undefined && { serverTiming })
```

newswires/client/src/context/fetchResults/fetchResults.ts
```diff
@@ -46,8 +56,40 @@ export const fetchResults = async ({
...
+
+	const serverTiming = extractServerTiming(response.headers);
+
 	return {
-		...parseResult.data,
-		results: parseResult.data.results.map(transformWireItemQueryResult),
+		data: {
+			...parseResult.data,
+			results: parseResult.data.results.map(transformWireItemQueryResult),
+		},
```
lindseydew, line 63: Think it would make sense to group these into request metadata: 
meta: {
		requestId,
		serverTiming,
	},

newswires/client/src/context/SearchContext.tsx
```diff
@@ -276,14 +276,16 @@ export function SearchContextProvider({ children }: PropsWithChildren) {
 			const startedAt = performance.now();
 			refreshesSinceLastTelemetrySend.current = 0;
 			fetchResults({ query: currentConfig.query, view: currentConfig.view })
-				.then((data) => {
+				.then(({ data, requestId, serverTiming }) => {
 					sendTelemetryEvent(
 						'NEWSWIRES_FETCHED_RESULTS',
 						createFetchedResultsTelemetryData({
```
lindseydew, line 282: Keep thinking this method could be worth a unit test at some point, but not necessarily in scope for this change

lindseydew (commented):
Nice looks like a really good initiative. Just a few minor coding style suggestions

newswires/client/src/context/fetchResults.ts
bryophyta: I was only planning to pass the total duration for now, but the reason for the specific format here is to match the 'timing-metric' format used by browsers: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Server-Timing#timing-metric

This allows the metrics to show up nicely in DevTools (screenshot in the PR description)

newswires/app/lib/RequestLoggingFilter.scala
bryophyta: The `db` metric in the unit tests is just there as an example of another possible metric, so that we can verify the function is picking out the `total` specifically, rather than other metrics which might be present in the header using the same format.

newswires/client/src/context/fetchResults.ts
bryophyta: > Is the idea that a new uuid is generated when it's called in application logic, and passed in explicitly for testing?

yep that was my reasoning 👍

I don't love mocking `uuidv4` for the tests, because I don't think we want to be too tied to it as an implementation detail, and also it seems conceivable that other internals might use `uuidv4` at some point, in which case they would also be affected. 

So I've opted to wrap the call in a separate `generateRequestId` function which we can mock instead. What do you think?

73c1f8c0

newswires/client/src/context/SearchContext.tsx
bryophyta: agree that it's simpler, but I think I'd prefer not using `&&` here, because technically this ends up evaluating to `...(false)` if `serverTiming === undefined`, which means that when we read it we need to remember that `{ ...false }` is equivalent to `{ ...{} }`, but I think the latter is more intuitive

newswires/app/lib/RequestLoggingFilter.scala
lindseydew: Cool, make sense. My only very minor nit to follow this would be for the first test to verify the behaviour would be to have a unit test for what the server will actually send, and perhaps renaming the second metric `db` to something like `test` or `dummy` metric, just because I got the impression from the tests that the intended server response would be 
`'server-timing': 'total;dur=123.45, db;dur=67.89'`

newswires/client/src/context/fetchResults/generateRequestId.ts
```diff
@@ -0,0 +1,8 @@
+import { v4 as uuidv4 } from 'uuid';
+
+/** Just a simple wrapper around uuidv4 to make it easier to mock in tests.
+ * Avoids the test setup needing to know about the internal implementation of fetchResults.
+ */
+export function generateRequestId(): string {
```
lindseydew, line 6: Nice, yeah I think this wrapper makes sense 👍

newswires/client/src/context/fetchResults/extractServerTiming.ts
```diff
@@ -0,0 +1,24 @@
+import { getErrorMessage } from '@guardian/libs';
+
+export function extractServerTiming(
+	headers: Headers | undefined,
+): number | undefined {
+	const serverTimingHeader = headers?.get('Server-Timing');
+	const serverTimingMatch = serverTimingHeader?.match(
+		/total;dur=(\d+(\.\d+)?)/,
+	);
+	try {
```
lindseydew, line 10: Small nit, 

```
  if(!serverTimingMatch) return undefined
  const serverTiming  = Number(serverTimingMatch[1])
  if (Number.isNaN(serverTiming)) return undefined
  return serverTiming
```

Might be slightly easier to parse

lindseydew (approved):
Looks great, great initiative 💯

PR #888, opened 2026-06-01
Update count query cap to 500

hk-15:
## What does this change?

Updates count query cap to 500 which ensures the load more button remains visible for longer. Before, the button disappeared after 3 clicks. 

## Introducing DB migrations?


lindseydew (approved):
LGTM 👍

PR #889, opened 2026-06-02
Remove the refresh button in the date picker

lindseydew:
## What does this change?
This updates the settings in the `EuiSuperDatePicker` so that the refresh button doesn't display because this was confusing to the end user. 

## Images

Before
image

After
image

## Introducing DB migrations?


bryophyta (commented):
fab! 🥳

andrew-nowak (approved):

PR #898, opened 2026-06-04
Add advance fixtures to Soccer Scores preset

bryophyta:
Advance fixtures (lists of upcoming football matches) were being excluded from all of the soccer presets. This adds them into the Soccer Scores preset (to parallel inclusion of similar fixtures in Rugby Scores).


hk-15 (approved):
LGTM!

PR #902, opened 2026-06-08
Improve documentation

hk-15:
## What does this change?

Using the [recommendations template](https://github.com/guardian/recommendations/blob/main/documentation.md#readme-template) and Copilot for a first pass, I have generated some improved documentation for this project.

It should cover all the basic points and make the tool easier to understand and get set up on for newcomers, with more information on scripts and commands that can be run.

There is some guidance on writing effect documentation here: https://github.com/guardian/recommendations/blob/main/documentation.md#readme-template

> Reviewers should be transparent about areas they found confusing or irrelevant. Identifying these areas in the documentation is the most helpful feedback an author can receive. It allows an author to consider their real audience and work with them to provide the right level of detail.


README.md
```diff
@@ -1,71 +1,146 @@
...
+- Ingesting wire content from multiple suppliers
+- Enriching and validating incoming records
+- Persisting content in PostgreSQL for filtering and search
+- Archiving raw payloads in S3 for audit and troubleshooting
+- Serving content via a Play backend and web UI
+
+External services and suppliers include:
+
+- Fingerpost feed integration (via SNS)
+- Reuters and AP feed integrations (via poller lambdas)
```
bryophyta:
```suggestion
- Copy filed by email from sports correspondents (ingested via AWS SES)
- Reuters and AP feed integrations (via poller lambdas) [Not currently used in PROD]
```

README.md
```diff
@@ -1,71 +1,146 @@
...
+- AWS managed services (Lambda, SQS, SNS, S3, RDS, EC2/ALB)
```
bryophyta: It feels slightly odd to me to include these AWS services in the list, insofar as the first two items on the list are sources of data, whereas this is more about general hosting details. 

Personally, I think we could probably drop the list of AWS services, but if we want to mention these, would it make sense to do it separately?

README.md
```diff
@@ -1,71 +1,146 @@
...
+
+The local environment checks in `scripts/check-requirements` expect:
 
-- [`poller-lambdas` README](poller-lambdas/README.md)
+- Java (see [.tool-versions](./.tool-versions)) with `JAVA_HOME` set
+- sbt
+- Node (see [.nvmrc](./.nvmrc))
+- Docker
+- nginx and dev-nginx
+- scala-cli
```
bryophyta: we should probably add `ssm` to this list https://github.com/guardian/ssm-scala

and it might be worth adding a link to dev-nginx too: https://github.com/guardian/dev-nginx

(we also need to update the `check-requirements` script to add those deps)

bryophyta (approved):
this looks like a great improvement to me -- the running instructions in particular are more concise and more readable 🙌 

i've added a couple of small comments but non-blocking 👍

README.md
```diff
@@ -135,3 +210,36 @@ graph TB
...
     class ASG,ALB compute
 ```
+
+### Key subprojects
+
+- [newswires](./newswires): Play backend and app packaging
+- [newswires/client](./newswires/client): React/Vite frontend, see [newswires/client/README.md](./newswires/client/README.md)
+- [ingestion-lambda](./ingestion-lambda): critical-path content processing + persistence, see [ingestion-lambda/README.md](ingestion-lambda/README.md)
+- [poller-lambdas](./poller-lambdas): supplier pollers + self-queueing mechanisms, see [poller-lambdas/README.md](poller-lambdas/README.md)
+- [fingerpost-queueing-lambda](./fingerpost-queueing-lambda): bridges Fingerpost SNS messages into ingestion queueing, see [fingerpost-queueing-lambda/README.md](fingerpost-queueing-lambda/README.md)
```
bryophyta: this reads slightly awkwardly to me.. I'd be tempted to say "receives messages from Fingerpost via SNS, then persists them to S3 and queues the items for processing by the ingestion lambda"? (though my version is a bit awkward too!)

README.md
hk-15: Good shout. Have done so 👍

PR #904, opened 2026-06-08
Add missing sports results to presets

hk-15:
## What does this change?

The cycling preset was missing results content from PA. This adds the general sport results category code to the list of PA sport codes to correct this. This change also adds missing results to the athletics preset.

## How to test

Missing results in cycling and athletics present in the presets on this branch, checked against PROD.


bryophyta (approved):
lgtm, nice!

PR #905, opened 2026-06-09
Exclude remaining media direct stories from presets where we're receiving an alternative feed

bryophyta:
## What does this change?

We're getting the new API feeds for all sports except Cricket and Racing, now. So we can filter out stories received via Media Direct from the following presets:

* Soccer scores
* Soccer tables
* Basketball scores
* Cycling
* ~Golf scores~ (see discussion below)
* Ice hockey
* Motor sport
* Rugby league
* Rugby union
* Rugby scores
* Tennis scores

We don't expect to find MD stories in any other presets (other than Cricket and Racing ones). Filtering these out helps us to preview the switch-off of the MD feeds later this month, while keeping all stories available via 'All' or 'All sports' presets, as a fallback in case there are any gaps in the new API feed.

Co-authored-by: hk-15 <168672047+hk-15@users.noreply.github.com>

## How to test

- [x] Run locally and compare the following two URLs. You should see some stories in CODE but none in local:
    - https://newswires.code.dev-gutools.co.uk/feed?guSourceFeed=PA+PA+SPORT+DATA&preset=soccer-scores
    - https://newswires.local.dev-gutools.co.uk/feed?guSourceFeed=PA+PA+SPORT+DATA&preset=soccer-scores

## Introducing DB migrations?

### DIFF
--- a/newswires/app/conf/SearchPresets.scala
+++ b/newswires/app/conf/SearchPresets.scala
@@ -882,8 +882,7 @@ object SearchPresets {
     SearchPreset(
       PA,
       searchTerms = Some(SingleTerm(SingleField("GOLF", Slug))),
-      categoryCodes = Some(CategoryCodesCondition(List("paCat:RSR"), SOME)),
-      guSourceFeedsExcl = SearchParams.mediaDirectSourceFeeds
+      categoryCodes = Some(CategoryCodesCondition(List("paCat:RSR"), SOME))
     ),
     SearchPreset(
       AP,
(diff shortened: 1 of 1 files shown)


lindseydew (approved):
LGTM 👍

blishen:
Thank you for this, I've just run through them all in CODE vs PROD and I think it looks good with the exception of Golf Scores where I don't see an equivalent PA_API story for things like: https://newswires.gutools.co.uk/item/2997221 I am reporting to Fingerpost but in the meanwhile if we could **allow** media direct content in that preset as you suggested I would be grateful.

PR #906, opened 2026-06-09
Add badge to media direct stories

bryophyta:
Co-authored-by: Holly Knox <168672047+hk-15@users.noreply.github.com>

## What does this change?

Add a 'MD' badge to stories that have come from Media Direct (legacy PA delivery source), to help us identify these stories as they're being phased out.

The code follows the pattern set for other badges (`lead` and `alert`): enhance the wire item in the client on ingestion to work out if `isMediaDirectItem`, then use this in subsequent rendering code.

## How to test

- [x] Run storybook and check out:
    - [x] http://localhost:6006/?path=/story/components-wireitemlist--with-data-formatting-and-media-direct-item-badge
    - [x] http://localhost:6006/?path=/story/components-item--with-multiple-labels
- [x] Run app locally and verify that badges appear only on media direct stories 

## Images

image

## Introducing DB migrations?

### DIFF
--- a/newswires/client/src/WireItemList.tsx
+++ b/newswires/client/src/WireItemList.tsx
@@ -21,7 +21,11 @@ import { Link } from './Link.tsx';
 import type { WireData } from './sharedTypes.ts';
 import { SupplierBadge } from './SupplierBadge.tsx';
 import { ToolSendReport } from './ToolsConnection.tsx';
-import { AlertLabel, LeadLabel } from './WireItemLabel.tsx';
+import {
+	AlertLabel,
+	LeadLabel,
+	MediaDirectItemLabel,
+} from './WireItemLabel.tsx';
 
 export const WireItemList = ({
 	wires,
@@ -201,6 +205,7 @@ const WirePreviewCard = ({
 		hasDataFormatting,
 		isAlert,
 		isLead,
+		isMediaDirectItem,
 	} = wire;
 
 	const ref = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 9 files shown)


newswires/client/src/utils/contentHelpers.ts
```diff
@@ -1,7 +1,19 @@
-import type { WireData } from '../sharedTypes.ts';
+import type { WireData, WireDataFromAPI } from '../sharedTypes.ts';
 
 export const isAlert = (content: WireData['content']) =>
 	content.type === 'text' && content.profile === 'alert';
 
 export const isLead = (content: WireData['content']) =>
 	content.type === 'composite' && content.profile === 'story';
+
+const mediaDirectSourceFeeds = [
```
bryophyta, line 9: this list is duplicated in the Scala code, but this whole bit of code is a) temporary, and b) not expected to change much while it's here, so trying to pass the list of sources from the server didn't seem worthwhile here.

newswires/client/src/WireItemList.tsx
```diff
@@ -386,8 +391,16 @@ const WirePreviewCard = ({
 						grid-area: badges;
 						justify-self: end;
 						align-self: flex-start;
+						display: flex;
+						gap: 0.3rem;
```
bryophyta, line 395: this is a hard-coded gap value which corresponds with the grid-column-gap for the card as a whole

lindseydew (approved):
Looks really good 👍

PR #911, opened 2026-06-10
ADRs for post-ingestion enhancements and for preset definitions

bryophyta:
## What does this change?

Add retrospective ADRs to capture decisions we've made about how to approach post-ingestion enhancement of story metadata, and pre-computation of search presets.

## How to test

- Do they make sense?
- Do they seem accurate reflections of the decisions we've taken, and the reasons for these decisions?

## Introducing DB migrations?

### DIFF
--- a/docs/adrs/03-post-ingestion-enhancement.md
+++ b/docs/adrs/03-post-ingestion-enhancement.md
@@ -34,7 +34,7 @@ What about the user who has P2 open?
     1. At the top, in which case it will potentially be out of chronological order with other items (sorted by ingestion time).
     2. In its chronological place, in which case it might not be at the top of the list, or even visible in the user's viewport at the time it loads.
 
-Neither (2.1) nor (2.2) feels ideal, and either option would also introduce a difference between the experience of P1 and P2 which would be impossible for end-users to diagnose.
+Neither (2.i) nor (2.ii) feels ideal, and either option would also introduce a difference between the experience of P1 and P2 which would be impossible for end-users to diagnose.
 
 ## Decision
 
--- a/docs/adrs/03-post-ingestion-enhancement.md
+++ b/docs/adrs/03-post-ingestion-enhancement.md
@@ -31,10 +31,10 @@ What about the user who has P2 open?
 
 1. We could just say they don't get to see it in their feed, although it will naturally show if they freshly load P2 after T2, but this seems like poor service.
 2. We could load S into their feed at T2, but then where should we place it in their feed?
-    i. At the top, in which case it will potentially be out of chronological order with other items (sorted by ingestion time).
-    ii. In its chronological place, in which case it might not be at the top of the list, or even visible in the user's viewport at the time it loads.
+    1. At the top, in which case it will potentially be out of chronological order with other items (sorted by ingestion time).
+    2. In its chronological place, in which case it might not be at the top of the list, or even visible in the user's viewport at the time it loads.
 
-Neither (2.i) nor (2.ii) feels ideal, and either option would also introduce a difference between the experience of P1 and P2 which would be impossible for end-users to diagnose.
+Neither (2.1) nor (2.2) feels ideal, and either option would also introduce a difference between the experience of P1 and P2 which would be impossible for end-users to diagnose.
 
 ## Decision
 
--- a/docs/adrs/03-post-ingestion-enhancement.md
+++ b/docs/adrs/03-post-ingestion-enhancement.md
@@ -4,7 +4,7 @@
 
 One advantage of having our own persistent storage for wire items (see [ADR 01](./01-newswires-overview.md)) is that it gives us more opportunities to enhance the metadata that's available for a given wire item. Example use-cases include:
 
-- Mapping Fingerpost's 'source feed' field [to our 'supplier' field](https://github.com/guardian/newswires/blob/9ede2e190239f087d91c7539ab6f983f1f1a1fd3/ingestion-lambda/src/processContentObject.ts#L254)
+- Mapping Fingerpost's 'source feed' field [to our 'supplier' field](https://github.com/guardian/newswires/blob/2c9a5e2c735f4455195b8d2fdd290bb18a89e6ca/ingestion-lambda/src/processContentObject.ts#L254)
 - Inferring whether a story is pertinent to the UK desk (https://github.com/guardian/newswires/pull/235)
 - Adding a TS vector index to enable better full text search
 
--- a/docs/adrs/03-post-ingestion-enhancement.md
+++ b/docs/adrs/03-post-ingestion-enhancement.md
@@ -1,4 +1,4 @@
-# ADR 02: Post-ingestion enhancement of wire items
+# ADR 03: Post-ingestion enhancement of wire items
 
 ## Context
 
@@ -10,7 +10,7 @@ One advantage of having our own persistent storage for wire items (see [ADR 01](
 
 Any such enhancement needs to either:
 
-1. blocks the story from being initially persisted in our database, or
+1. block the story from being initially persisted in our database, or
 2. take place in a non-blocking way, either asynchronously or triggered after ingestion
 
 ## Options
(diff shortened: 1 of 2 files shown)


docs/adrs/03-post-ingestion-enhancement.md
```diff
@@ -0,0 +1,43 @@
...
+
+One advantage of having our own persistent storage for wire items (see [ADR 01](./01-newswires-overview.md)) is that it gives us more opportunities to enhance the metadata that's available for a given wire item. Example use-cases include:
+
+- Mapping Fingerpost's 'source feed' field [to our 'supplier' field](https://github.com/guardian/newswires/blob/9ede2e190239f087d91c7539ab6f983f1f1a1fd3/ingestion-lambda/src/processContentObject.ts#L254)
+- Inferring whether a story is pertinent to the UK desk (https://github.com/guardian/newswires/pull/235)
+- Adding a TS vector index to enable better full text search
+
+Any such enhancement needs to either:
+
+1. blocks the story from being initially persisted in our database, or
```
hk-15: nit:
```suggestion
1. block the story from being initially persisted in our database, or
```

docs/adrs/03-post-ingestion-enhancement.md
```diff
@@ -0,0 +1,43 @@
...
+Post-ingestion enhancement introduces a significant complication, however. Imagine that the enhancement of a story S affects how S is categorised by preset or user-defined filters. On ingestion at time T1, it fits the criteria for preset P1, and then some time later at T2 an enhancement is added which means it also fits the criteria for P2. Now imagine that users are watching feeds of P1 and P2.
+
+The user who has P1 open from T0 to T2 will see story S show up just after T1.
+
+What about the user who has P2 open?
+
+a. We could just say they don't get to see it in their feed, although it will naturally show if they freshly load P2 after T2, but this seems like poor service.
+b. We could load S into their feed at T2, but then where should we place it in their feed?
+i. At the top, in which case it will potentially be out of chronological order with other items (sorted by ingestion time).
+ii. In its chronological place, in which case it might not be at the top of the list, or even visible in the user's viewport at the time it loads.
```
hk-15: nit: perhaps more readable as a numbered list so the formatting applies?

docs/adrs/03-post-ingestion-enhancement.md
```diff
@@ -0,0 +1,43 @@
...
+
+Neither (b.i) nor (b.ii) feels ideal, and either option would also introduce a difference between the experience of P1 and P2 which would be impossible for end-users to diagnose.
+
+## Decision
+
+We have decided not to do asynchronous or post-ingestion enhancements of wire items if the enhancement concerns attributes which we user for constructing live feeds. Given that, as per [the ADR on Filter and Search](./02-filter-and-search.md), we are aiming to model everything via the idea of a unified filter/search view presented to the user, this effectively means that we aren't doing asynchronous enhancement at the moment. However, there's no reason in principle why we couldn't add asynchronous enhancement in the future if we add other discovery/browsing features ('trending topics' might be a good example).
```
hk-15: nit:
```suggestion
We have decided not to do asynchronous or post-ingestion enhancements of wire items if the enhancement concerns attributes which we use for constructing live feeds. Given that, as per [the ADR on Filter and Search](./02-filter-and-search.md), we are aiming to model everything via the idea of a unified filter/search view presented to the user, this effectively means that we aren't doing asynchronous enhancement at the moment. However, there's no reason in principle why we couldn't add asynchronous enhancement in the future if we add other discovery/browsing features ('trending topics' might be a good example).
```

docs/adrs/03-post-ingestion-enhancement.md
```diff
@@ -0,0 +1,43 @@
+# ADR 02: Post-ingestion enhancement of wire items
+
+## Context
+
+One advantage of having our own persistent storage for wire items (see [ADR 01](./01-newswires-overview.md)) is that it gives us more opportunities to enhance the metadata that's available for a given wire item. Example use-cases include:
+
+- Mapping Fingerpost's 'source feed' field [to our 'supplier' field](https://github.com/guardian/newswires/blob/9ede2e190239f087d91c7539ab6f983f1f1a1fd3/ingestion-lambda/src/processContentObject.ts#L254)
```
hk-15: This link doesn't appear to work?

docs/adrs/04-precomputation-of-presets.md
```diff
@@ -0,0 +1,85 @@
...
+
+In order to create a more familiar and easily adoptable experience for users, we plan to recreate many of the lists/feeds that were provided by the Fingerpost UI (e.g. 'UK News', 'Cricket scores'). Under the hood, these will be saved searches which can be combined with other user-defined filters and search terms. We've sometimes referred to these internally as 'buckets', but in the UI we've gone for 'presets'.
+
+The searches underlying the familiar presets are often very complex (as they were in the Fingerpost implementation) because different agencies provide different metadata and none of them map precisely onto the categories our users are interested in.As an example, we might need to add inclusions and exclusions based on category code and 'slug' text. For example, in order to differentiate Rugby League from Rugby Union, and to differentiate rugby _stories_ from rugby _results_, we need to look for `"RUGBYL" / "RUGBYU" / "RUGBY UNION"` in the slug, and exclude any of a number of category codes which are applied to results only:
```
hk-15: nit (missing space):
```suggestion
The searches underlying the familiar presets are often very complex (as they were in the Fingerpost implementation) because different agencies provide different metadata and none of them map precisely onto the categories our users are interested in. As an example, we might need to add inclusions and exclusions based on category code and 'slug' text. For example, in order to differentiate Rugby League from Rugby Union, and to differentiate rugby _stories_ from rugby _results_, we need to look for `"RUGBYL" / "RUGBYU" / "RUGBY UNION"` in the slug, and exclude any of a number of category codes which are applied to results only:
```

docs/adrs/03-post-ingestion-enhancement.md
```diff
@@ -0,0 +1,43 @@
+# ADR 02: Post-ingestion enhancement of wire items
```
hk-15: Small change:
```suggestion
# ADR 03: Post-ingestion enhancement of wire items
```

docs/adrs/04-precomputation-of-presets.md
```diff
@@ -0,0 +1,85 @@
+# Precomputation of search presets
```
hk-15: Nit, but for consistency?
```suggestion
# ADR 04: Precomputation of search presets
```

docs/adrs/04-precomputation-of-presets.md
```diff
@@ -0,0 +1,85 @@
...
+```sql
+SELECT i.headline
+FROM item i
+JOIN item_presets ip ON ip.item_id = i.id
+WHERE ip.preset_id = $1
+ORDER BY i.first_created DESC
+LIMIT 100;
+```
+
+There are good reasons to think that this would be quicker to execute at query time than the complex queries we currently have. If it worked as expected, then it would also unlock the possibility of combining presets in a given user query (e.g. "all sports stories that aren't in the cricket presets").
```
hk-15: Really minor nit:
```suggestion
There are good reasons to think that this would be quicker to execute at query time than the complex queries we currently have. If it worked as expected, then it would also unlock the possibility of combining presets in a given user query (e.g. 'all sports stories that aren't in the cricket presets').
```

hk-15 (approved):
This is great and very clearly explained – I have learnt a lot from reading it! I have a few non-blocking comments, mostly nitpicks and not essential.

docs/adrs/03-post-ingestion-enhancement.md
bryophyta: good catch, thanks! fixed now 👍

docs/adrs/03-post-ingestion-enhancement.md
bryophyta: thanks! for some reason I thought it would format these as I wanted, but TIL!

PR #913, opened 2026-06-11
Show S3 writes in diagram, and add dot-copy

bryophyta:
We haven't updated the diagram for a while, so it was missing the fact that the ingestion lambda now *reads* from S3, and upstream sources write to it. It was also missing the 'dot-copy' feed via SES.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,11 +143,12 @@ CI-style workspace commands can also be run by appending `:ci` to any of the abo
 ### High-level architecture
 
 ```mermaid
-graph TB
+graph LR
     %% External Sources
     Reuters[Reuters Feed]
     AP[AP Feed]
     Fingerpost[Fingerpost Feed]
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -144,11 +143,12 @@ CI-style workspace commands can also be run by appending `:ci` to any of the abo
 ### High-level architecture
 
 ```mermaid
-graph TB
+graph LR
     %% External Sources
     Reuters[Reuters Feed]
     AP[AP Feed]
     Fingerpost[Fingerpost Feed]
+    Correspondents[Correspondents ('dot-copy')]
```
andrew-nowak: github errors trying to render the diagram, pointing to this line. I guess it doesn't like the brackets 

Image

bryophyta: thanks! thought I was making a safe tweak there but nope! I've fixed now 👍

andrew-nowak (approved):

PR #915, opened 2026-06-15
Ensure ticker can be opened from the side nav for AAP and Minor Agencies

hk-15:
## What does this change?

When a user opened a ticker from the side nav for AAP, it returned no search results because the supplier label (which contains parentheses and a special character) was being used for the search.

This PR fixes the problem by ensuring that the supplier name is used for the search when the ticker is opened. This fix will also ensure that the Minor Agencies ticker opens correctly.

## How to test

For AAP and Minor Agencies: verify that opening a ticker using the side nav yields the same result as opening one via the `New Ticker` button in the top right corner. Test opening tickers for all suppliers and combinations of a few suppliers to check that no existing functionality has been affected.

### DIFF
--- a/newswires/client/src/SideNav/SideNav.tsx
+++ b/newswires/client/src/SideNav/SideNav.tsx
@@ -103,6 +103,7 @@ export const SideNav = ({
 	const suppliers = useMemo(
 		() => [
 			{
+				name: "",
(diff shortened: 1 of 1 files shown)


bryophyta (approved):
nice! this lgtm!

PR #925, opened 2026-06-26
Fix badge alignment and Storybook config

bryophyta:
I was puzzled why the alignment of badges in the WireItemList seemed to be as intended in the Storybook, but not in the app. The Storybook stories weren't receiving the global font styling that we apply to the app.

This PR:

1. Adds the same global styles to all Storybook stories as we add to the app, to ensure a more consistent testing environment
2. Removes the global line-height rule which seemed to be responsible for the unexpected alignment in WireItemList

|Before|After|
|-|-|
|image|image|


hk-15 (commented):
Looks good 👏

hk-15 (approved):

PR #926, opened 2026-06-26
Upgrade EUI to 116.x

bryophyta:
## What does this change?

Bump EUI to v116. There are a few API changes which don't seem to create visual differences from what I can see. There are also a few tweaks to the default colour scheme. The latter don't seem super significant, and we've been using the default so far, so I think that's an indication that we should continue here. If we want to set our own custom theme at some point then we can do so.

## Introducing DB migrations?


newswires/client/src/icons.ts
andrew-nowak, line 1: Do icons now work properly out of the box?? 🥳

andrew-nowak (approved):

newswires/client/src/icons.ts
bryophyta: yes, it seems so!!

PR #927, opened 2026-06-26
Wire card layout fixes

bryophyta:
## What does this change?

Over the last couple of months we've added new types of labels to the wire item list cards, e.g. 'Lead' and 'Alert' labels.

These have sat alongside the supplier badges and are conceptually and visually very similar, but there have been some alignment issues between them. #925 introduced one fix for the alignment issues, but I realised after merging that it had a bigger impact on the site than intended (reducing the line height across the app by quite a bit).

This PR attempts a more robust fix by standardising and simplifying the rendering of supplier and other labels in the list.

### Extend WireItemLabel for suppliers

Supplier badges have always used `EuiBadge` under the hood in this project. A `WireItemLabel` component was introduced a little while ago to allow for the 'lead' and 'alert' labels in the wire item list, which couldn't easily be done via `EuiBadge` without adding fragile style overrides.

This branch uses WireItemLabel to render supplier badges as well. This means that the spacing and typography for the supplier badge and other labels is more consistent, and makes it easier to ensure that they are aligned to each other properly. It also has the advantage of insulating us from some up-coming changes to badge styling in ElasticUI which we are quite noticeable.

### Simplify the grid layout for wire cards

Having made the above change, and given that we're using labels/badges more often over time, it felt sensible to group them all into the same grid area, which makes the layout a bit simpler to reason about.

### Remove grid-row-gap from the card grid

The non-compact card layout is complex, because it's possible for either (or both) of the headline and subhead to wrap onto multiple lines, and at the same time we want:

1. for the time and the supplier badge to stack vertically, to the side of the headline & subheading, and
2. to reduce unnecessary vertical whitespace

To try and achieve a decent balance while just using CSS for layout, we have the badges area overlap with both the title and subhead rows. Previously there was a `grid-row-gap` applied to the grid as a whole, to provide vertical spacing between areas. However, the row gap ends up stacking in a way that pushes the subheading below the badges when the headline is short. I've replaced the row gap with margin on the bottom of the elements on the top row, which means that the badges and the subheading start at the same point if the headline is short and doesn't wrap.

image

|Before|After|
|-|-|
|image|image|
|image|image|
|image|image|

The card layout it still a bit complicated, and hopefully we can find some more simplifications at some point.

## How to test

- [x] `npm run storybook`
- [x] do the stories look okay?
- [x] are there any important edge cases not covered by the existing stories?

## Introducing DB migrations?


newswires/client/src/WireItemList.stories.tsx
```diff
@@ -206,11 +206,11 @@ export const WithDateOlderThan24Hours: Story = {
 		wires: [
 			{
 				...sampleItemData,
-				ingestedAt: new Date(Date.now() - 48 * 60 * 60 * 1000).toISOString(), // 48 hours ago
```
bryophyta, line 209: pass-by fix -- the card uses `ingestedAtMoment` now, so setting `ingestedAt` wasn't doing anything

andrew-nowak (approved):

PR #928, opened 2026-06-26
Add developer policy for local running against CODE db

bryophyta:
## What does this change?

Adds a developer policy with all (and hopefully only) the permissions you need in order to run the app locally against the CODE database, e.g. using `./scripts/start --use-CODE`.

I also had to upgrade GuCDK in order to get the right constructs, so there are a couple of changes in this branch to accommodate that. They could be split out into a separate PR if distracting here.

## How to test

- [x] Deploy branch to CODE
- [x] Grant yourself [DeveloperPolicyGrants.newswiresLocalRunAgainstCodeDb](https://github.com/guardian/janus/pull/5622/changes#diff-d974b4aad3f15e98ace51a3732bf68e17cccd5e30d6b503b51f1bd7a10639719L270) access via Janus
- [x] Get credentials for this policy from Janus
- [x] Run `./scripts/start --use-CODE` to start Newswires locally; everything should work as it does when you have full developer credentials available locally

## Introducing DB migrations?

### DIFF
--- a/cdk/lib/newswires.ts
+++ b/cdk/lib/newswires.ts
@@ -71,6 +71,7 @@ import { POLLERS_CONFIG } from 'newswires-shared/pollers';
 import { appName, LAMBDA_ARCHITECTURE, LAMBDA_RUNTIME } from './constants';
 import { GuDatabase } from './constructs/database';
 import { PollerLambda } from './constructs/pollerLambda';
+import { createLocalRunDeveloperPolicy } from './local-run-developer-policy';
 
 export type NewswiresProps = GuStackProps & {
 	sourceQueue: Queue;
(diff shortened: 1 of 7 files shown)


cdk/lib/newswires.ts
```diff
@@ -579,6 +581,19 @@ export class Newswires extends GuStack {
 			database.accessSecurityGroup,
 		);
 
+		// Developer policy allowing engineers to run the app locally against the
+		// CODE database (./scripts/start --use-CODE). No-op outside the CODE stack (and included in TEST so we get the snapshot).
```
lindseydew, line 585: 👍

lindseydew (approved):
LGTM 👍  nice work

PR #930, opened 2026-06-29
Upgrade Storybook to v10

bryophyta:
Upgrade Storybook to v10. Requires a little bit of work because of the breaking change in support for CommonJS modules.

### DIFF
--- a/newswires/client/.storybook/main.ts
+++ b/newswires/client/.storybook/main.ts
@@ -1,6 +1,12 @@
 import type { StorybookConfig } from '@storybook/react-vite';
 
-import { join, dirname } from 'path';
+import { dirname, join } from 'path';
+
+import { createRequire } from 'node:module';
+import { fileURLToPath } from 'node:url';
+
+const require = createRequire(import.meta.url);
+const __dirname = dirname(fileURLToPath(import.meta.url));
 
 /**
  * This function is used to resolve the absolute path of a package.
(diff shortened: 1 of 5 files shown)


newswires/client/.storybook/main.ts
```diff
@@ -1,6 +1,12 @@
 import type { StorybookConfig } from '@storybook/react-vite';
 
-import { join, dirname } from 'path';
+import { dirname, join } from 'path';
+
+import { createRequire } from 'node:module';
+import { fileURLToPath } from 'node:url';
+
+const require = createRequire(import.meta.url);
```
bryophyta, line 8: Storybook has moved to ESM only so we need this to allow `require` to work below

newswires/client/src/sharedTypes.ts
```diff
@@ -156,7 +156,7 @@ export type WiresQueryData = {
 	};
 };
 
-export const isValidDateValue = (value: string): value is EuiDateString =>
```
bryophyta, line 159: Might not strictly be necessary for this branch -- we shouldn't be touching TS? -- but future versions of TS flag this as a circular declaration and I was getting an error when typechecking on this branch earlier so might as well get the fix in now..?

newswires/client/package.json
```diff
@@ -33,11 +32,9 @@
 		"zod": "4.0.5"
 	},
 	"devDependencies": {
-		"@guardian/prettier": "^8.0.1",
-		"@storybook/addon-onboarding": "9.1.17",
```
bryophyta, line 37: we don't need the onboarding plugin or the chromatic plugin as far as I can tell

andrew-nowak (approved):

PR #931, opened 2026-06-30
Convert from using Jest to using Vitest

bryophyta:
## What does this change?

We've had a few issues recently when upgrading various packages, due to limitations in Jest's support for ESM and TypeScript.

Vitest has been around for a while now, pitched as a Jest-compatible test runner which provides greater speed and compatibility with Vite (which we're using for the client dev server and build). It also has built-in support for ESM and TypeScript, meaning that we can drop `ts-jest` as well as Jest.

From local testing, it looks to be about twice as fast as Jest, although that's not a huge consideration here as the test suite doesn't take very long to run (~9-11s with Jest) and we're using Lage locally which caches test runs. But good to confirm that it moves us in the right direction.

Finally, this switch allows us to drop 293 packages from our `node_modules`, while only adding 30 news ones, compared with Jest. Quality matters as well as quantity, and I haven't done a proper audit of the packages that are added and removed, but my sense is that Vitest is generally working with more 'modern' dependencies. The numbers are quite compelling in themselves, though. 

## How to test

- [x] Do all the tests still pass?
- [x] I added a failing assertion to each of the 34 `*.test.{ts,tsx}` files in the repo and verified that this generated 34 failures. This gives me confidence that the tests aren't just passing because they're being accidentally skipped by the new config.

## Introducing DB migrations?


vitest.config.ts
```diff
@@ -0,0 +1,34 @@
...
+	test: {
+		name,
+		root: resolve(import.meta.dirname, name),
+		environment: 'node' as const,
+		globals: true,
+		include: ['**/*.test.ts'],
+		setupFiles: ['./vitest.setup.ts'],
+		passWithNoTests: true,
+	},
+});
```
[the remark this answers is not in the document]
bryophyta: yep, we don't have tests for these so have deleted them from the config; I forgot we didn't have tests for every npm workspace in this repo

package.json
```diff
@@ -14,8 +14,8 @@
 		"recomputation-lambda"
 	],
 	"scripts": {
-		"test": "lage test",
-		"test:ci": "npm run test --workspaces --if-present",
+		"test": "vitest run",
```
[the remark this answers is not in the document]
bryophyta: I don't think this is true

newswires/client/src/context/SearchContext.test.tsx
```diff
@@ -90,6 +88,8 @@ describe('SearchContext', () => {
 			contextRef.current?.handleEnterQuery(q);
 		});
 
+		await flushPendingPromises();
```
bryophyta, line 91: After removing the `disableLogs()` setup from `beforeEach`, it surfaced a warning message detecting changes to the state outside of `act()`. Adding this flush fixes that issue.

newswires/client/src/context/SearchContext.test.tsx
```diff
@@ -50,10 +49,9 @@ describe('SearchContext', () => {
 	};
 
 	beforeEach(() => {
-		jest.clearAllMocks();
+		vi.clearAllMocks();
 		localStorage.clear();
-		disableLogs();
```
bryophyta, line 55: A couple of warnings were being suppressed by this, and the client code doesn't do much of its own logging, so I thought it better to allow logging for all the tests in `newswires/client`, except for the moment.js warnings which we get when we intentionally pass an invalid input. I've added a more targeted helper to suppress just those expected and unhelpful log messages.

poller-lambdas/src/pollers/ap/apPoller.test.ts
```diff
@@ -29,12 +30,11 @@ describe('apPoller', () => {
 	const input = 'https://api.ap.org/media/v/content/feed?page=1';
 
 	beforeEach(() => {
-		jest.clearAllMocks();
-		jest.spyOn(console, 'log').mockImplementation(() => {});
```
bryophyta, line 33: I've set `silent: "passing-only"` in the root config for the repo now so we don't need to create logging mocks here anymore

shared/package.json
```diff
@@ -7,13 +7,10 @@
 	},
 	"scripts": {
 		"typecheck": "tsc -noEmit",
-		"pretest": "cd ../ && ./scripts/setup-test-env.sh",
```
bryophyta, line 10: I couldn't find a need for this. I tried deleting the test docker image and then running `npm run test --workspace shared` and everything ran fine

andrew-nowak (approved):

PR #946, opened 2026-07-09
Remove media direct exclusions from search presets

hk-15:
## What does this change?
Now that PA have phased out Media Direct and we have stopped receiving stories from this source, we need to clean up source feed logic that is no longer needed.

This PR reverts changes made in https://github.com/guardian/newswires/pull/905, as well as small changes in https://github.com/guardian/newswires/pull/916 and https://github.com/guardian/newswires/pull/923, all of which added exclusions to search presets to filter out Media Direct stories.

## How to test
Go to Cricket Scores preset, change date range to end at 10pm on 25 June and some MD stories should now be visible in the preset. This can be verified by checking the JSON of the top item, in which "guSourceFeed": "PA DATA FORMATTING" is present.

### DIFF
--- a/newswires/app/conf/SearchPresets.scala
+++ b/newswires/app/conf/SearchPresets.scala
@@ -3,7 +3,7 @@ package conf
 import conf.SearchField.{BodyText, Headline, Slug}
 import conf.SearchTerm.SingleField
 import conf.Suppliers._
-import models.{FilterParams, SearchParams}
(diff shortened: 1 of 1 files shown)


bryophyta (approved):
code changes look great, and really helpful write-up of how to test! I couldn't actually test with this because all the MD stories have been removed from the db now but it's really helpful documentation to have 🤩

PR #947, opened 2026-07-09
Hk/remove md switch

hk-15:
## What does this change?

Now that PA have phased out Media Direct and we have stopped receiving stories from this source, we need to clean up source feed logic that is no longer needed.

Remove the switch which was used to hide 'media direct' source feeds, by reverting changes made in PR https://github.com/guardian/newswires/pull/901.


bryophyta (approved):
lgtm!

PR #958, opened 2026-07-17
Add tagging for potential obituary-related phrases

bryophyta:
## What does this change?

Adds a `gu:obits` category code to stories on ingestion, based on matching against some phrases that are commonly used when announcing obituaries.

## How to test

- [x] Deploy to CODE, check that the tag is being applied sensibly
- [x] Leave the branch up for a few days and confirm that it's not making a noticeable difference to ingestion latency:

image

## Introducing DB migrations?

### DIFF
--- a/ingestion-lambda/src/categoryCodes.ts
+++ b/ingestion-lambda/src/categoryCodes.ts
@@ -243,6 +243,19 @@ export function inferGBCategoryFromText(content: string | undefined): string[] {
 	); /** @todo we should be able to remove the type predicate after we upgrade TS to 5.6 */
 }
 
+export function inferObituaryCategoryFromText(
+	content: string | undefined,
+): string[] {
+	if (!content) {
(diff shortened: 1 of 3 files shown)


ingestion-lambda/src/categoryCodes.ts
```diff
@@ -243,6 +243,19 @@ export function inferGBCategoryFromText(content: string | undefined): string[] {
...
 
+export function inferObituaryCategoryFromText(
+	content: string | undefined,
+): string[] {
+	if (!content) {
+		return [];
+	}
+
+	const hasObituaryPhrase =
+		/has died|died at age|died aged|dies at age|dies aged/i.test(content);
```
junyuanxue: Personally I'd pull this out into a constant but this works!

junyuanxue (approved):
LGTM!

PR #961, opened 2026-07-23
Add strictPort to vite config

junyuanxue:
## What does this change?

I was running `newswires` locally with port 5173 already occupied in the background, as a result I was served a blank page when going to https://newswires.local.dev-gutools.co.uk.

Port 5173 is required for nginx-mapping.yaml. `strictPort` makes a duplicate/stray instance fail loudly instead of silently binding to a different port.


hk-15 (approved):
Makes sense!

PR #962, opened 2026-07-23
Fix search empty state layout

junyuanxue:
## What does this change?

Search empty state is looking a bit broken at the moment. This PR fixes it (via some brute-force CSS overrides, lol)

### Before:
Screenshot 2026-07-23 at 17 36 21

### After:
Screenshot 2026-07-23 at 17 36 31

### DIFF
--- a/newswires/client/src/Feed.tsx
+++ b/newswires/client/src/Feed.tsx
@@ -147,7 +147,7 @@ export const Feed = ({ containerRef, setSideNavIsOpen }: FeedProps) => {
 			)}
 			{(status == 'success' || status == 'offline') &&
 				queryData.results.length === 0 && (
-					<>
+					<EuiFlexGroup direction="column">
 						<EuiFlexGroup>
 							<EuiFlexItem style={{ flex: 1 }}></EuiFlexItem>
 							<EuiFlexItem grow={false}>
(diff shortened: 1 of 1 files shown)


bryophyta (commented):
This is a good shout! I'm a bit wary of the margin override though.. I guess the root cause of it looking dodgy at the moment is that we're using the `SearchSummary` component for both the empty prompt and for the success case, and the success case gets looked at a lot more often which has led the empty case to drift over time.

Given this, it feels like an `!important` override is especially likely to get out of sync at a later date. Would it make sense to refactor the `SearchSummary` to be a bit more agnostic between the two locations instead? 

(If that feels like too much of a side quest though then please do say so!)

newswires/client/src/Feed.tsx
```diff
@@ -147,26 +151,33 @@ export const Feed = ({ containerRef, setSideNavIsOpen }: FeedProps) => {
...
+						>
+							<EuiText size="s">
+								<h2>No results match your search criteria</h2>
+							</EuiText>
+							<SearchSummary setSideNavIsOpen={setSideNavIsOpen} />
+							<EuiText color="subdued" component="p">
+								Try another search or reset filters.
+							</EuiText>
+						</EuiFlexGroup>
+					</EuiFlexGroup>
```
junyuanxue, line 180: @bryophyta (Replying to your comment here so we can 🧵 more easily.) I definitely did not like the `!important` at all 😂  I think the main issue was the h2 and p wrapped inside `EuiEmptyPrompt` had some extra styling on top, resulting in `SearchSummary` (which is also using an h2) looking very different in the empty state than the regular search result. One other option is to wrap this inside `EuiFlexGroup` - it'd be less semantic but possibly cleaner:
Image

bryophyta (approved):
nice! this looks so much cleaner 👍

PR #973, opened 2026-08-03
Clean up docker state before running

lindseydew:
## What does this change?

This runs a step to `remove orphans` in the test setup script. Occasionally I've experienced prehook commit fails and needed to run this step manually in order to restore the state of docker. This automates that step
## Introducing DB migrations?


hk-15 (approved):
Great! Thank you for your service 🙏

